PR #2, opened 2016-02-04
fixing tests

mm-binary:


MANIFEST
```diff
@@ -9,6 +9,7 @@ lib/Quant/Framework/Utils/Types.pm
 LICENSE
 Makefile.PL
 MANIFEST			This list of files
+Quant-Framework-0.01.tar.gz
```
tfoertsch: ???

Makefile.PL
```diff
@@ -29,6 +28,7 @@ WriteMakefile(
         'Test::More' => 0,
         'Test::Exception' => 0,
         'Test::NoWarnings' => 0,
+        'Test::CheckManifest' => 1.29,
```
tfoertsch: You require this module but the related test has been removed. Did you move it anywhere else?

tfoertsch:
I guess this is going to fail on cpantesters because of missing dependencies. But let's see.

tfoertsch:
I think it's a good idea to add back the MANIFEST test because this time the manifest was wrong.

Makefile.PL
mm-binary: Test is removed because its not used in other similar repos.

MANIFEST
mm-binary: Removed.

mm-binary:
manifest checking is back and is passing on travis.

Makefile.PL
```diff
@@ -16,7 +16,6 @@ WriteMakefile(
     AUTHOR           => q{Binary.com <support@binary.com>},
     VERSION_FROM     => 'lib/Quant/Framework.pm',
     ABSTRACT_FROM    => 'lib/Quant/Framework.pm',
-    LICENSE          => 'artistic_2',
     PL_FILES         => {},
     MIN_PERL_VERSION => 5.006,
```
tfoertsch: do we really care?

PR #6, opened 2016-02-18
Moving Dividend module to Q::F

mm-binary:


MANIFEST
```diff
@@ -2,6 +2,7 @@
 Changes
```
jy-binary: don't commit MANIFEST file?

jy-binary: sorry ignore that

kavehmz:
@mm-binary  Can you please rewrite the module in Moo instead of Moose.

The change in code will be minimal but increase in speed between these two implementations are noticeable.

https://gist.github.com/kavehmz/e648e2301f02ebe407ff

kavehmz:
@mm-binary Please rewrite the code on Moo instead of Moose.

Change in the code will be minimal but gain in performance will be noticeable.

lib/Quant/Framework/Dividend.pm
```diff
@@ -0,0 +1,181 @@
...
+    is         => 'rw',
+    lazy_build => 1,
+);
+
+sub _build_document {
+    my $self = shift;
+
+    my $document = $self->chronicle_reader->get('dividends', $self->symbol);
+
+    if ($self->for_date and $self->for_date->datetime_iso8601 lt $document->{date}) {
```
basiliscos: I feel a bit uneasy, when it is dates are compared as strings. Why not compare `$date->epoch` ?

mm-binary: Done.

basiliscos: :+1:

PR #9, opened 2016-03-08
Moving Holiday to Q::F

mm-binary:


basiliscos:
Trivial, as soon, as I saw the Holiday module + it's tests a few days ago.

basiliscos:
Looks fine for me.

lib/Quant/Framework/Holiday.pm
```diff
@@ -0,0 +1,104 @@
...
+Updates the current holiday calendar with the new inserts.
+It trims the calendar by removing holiday before the recorded_date.
+
+=cut
+
+sub save {
+    my $self = shift;
+
+    my $cached_holidays = $self->chronicle_reader->get('holidays', 'holidays');
+    my %relevant_holidays = map { $_ => $cached_holidays->{$_} } grep { $_ >= $self->recorded_date->truncate_to_day->epoch } keys %$cached_holidays;
```
tfoertsch: I think it's more efficient to delete old holidays like this:

```perl
my $recorded_date = $self->recorded_date->truncate_to_day->epoch;
delete @{$cached_holidays}{grep { $_ < $recorded_date } keys %$cached_holidays};
```

Then you don't even need `%relevant_holidays`.

mm-binary: I agree. Problem is `$cached_holiday` is an array and in the code that follows we will need a hash as we will be querying based on the `epoch` of events. So I prefer to keep it this way.

PR #10, opened 2016-03-15
moving economic events to Q::F

mm-binary:


lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+#this sub needs to be removed as it is no loger used.
+#we use `get_latest_events_for_period` to read economic events.
+sub _build_document {
+    my $self = shift;
+
+    #document is an array of hash
+    #each hash represents a single economic event
+    return $self->chronicle_reader->get(EE, EE);
+}
+
```
basiliscos: If not `document` property nor `_build_document` are no longer used used, they, should be removed

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+has events => (
+    is         => 'ro',
+    lazy_build => 1,
+);
+
+sub _build_events {
+    my $self = shift;
+    return $self->document->{events};
+}
+
```
basiliscos: The `events` & friends should be removed too, bacaue they used `document` (no longre used). The direct asses is used for events property in line 234

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+# economic events to be recorded to chronicle after processing.
+# tentative events without release_date will not be including in this list.
+has _events => (
+    is => 'rw',
+);
+
+around _document_content => sub {
```
basiliscos: Don't use **around**, unless you have serious reason to do that. 

I think the more correct way to achieve what you want is :

1. In `Quant::Framework::Utils::MarketData` : rename `_document_content` to something like `default_metadata`.

```perl

has _default_metadata => (
    is         => 'ro',
    lazy_build => 1,
);

sub _build__default_metadata {
    my $self = shift;

    return {
        symbol => $self->symbol,
        date   => $self->recorded_date->datetime_iso8601,
    };
}
```

2. In `EconomicEventCalendar` just have usual `_document_content` with lazy builder like:

```perl
sub _build__document_content {
    my $self = shift;
    my $metadata = $self->_default_metadata;
    my %content = (%$metadata, _events => $self_events)

    return \%content;
}
```

AFAIK, **around**, is used for interception, and should be used with care.

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+Saves the calendar into Chronicle
+
+=cut
+
+sub save {
+    my $self = shift;
+
+    for (EE, EET) {
+        $self->chronicle_writer->set(EE, $_, {}) unless defined $self->chronicle_reader->get(EE, $_);
+    }
```
basiliscos: What is the sense of the `for` cycle to initialize redis with empty values? At the end of the function (which is always reached), it sets real (with content) values.

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+    # 2) tentative events that we do not care. (those that we don't add blockout times for them, we treat it as if they don't exist)
+    # 3) tentative events that we care. (with blockout time and release_date)
+    my @regular_events =
+        sort { $a->{release_date} <=> $b->{release_date} } grep { $_->{release_date} } @{$self->events};
+    $self->_events(\@regular_events);
+
+    return (
+        $self->chronicle_writer->set(EE, EET, $existing_tentatives,     $self->recorded_date),
+        $self->chronicle_writer->set(EE, EE,  $self->_document_content, $self->recorded_date));
+}
```
basiliscos: I think it would be better just have 2 instructions `_writer->set(...)` and empty return instead of return of the result of two save operations. It will be less confusing.

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+
+=cut
+
+sub update {
+    my ($self, $params) = @_;
+
+    my $existing_events = $self->chronicle_reader->get(EE, EE) || {};
+    my $tentative_events = $self->get_tentative_events || {};
+
+    croak "Specify a blackout start and end to update tentative event" unless ($params->{blankout} and $params->{blankout_end});
```
basiliscos: This check operates only on `params` and should be done earlier (i.e. befor you get anything  from redis)

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+    croak "could not find $params->{id} in tentative table" unless $tentative_events->{$params->{id}};
+
+    $params->{release_date} = int(($params->{blankout} + $params->{blankout_end}) / 2);
```
basiliscos: Should be commented / explained, why do you take the average here as the release date.

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+    } else {
+        push @{$existing_events->{events}}, $params;
+    }
+
+    $tentative_events->{$params->{id}} = {(%{$tentative_events->{$params->{id}}}, %$params)};
+
+    return (
+        $self->chronicle_writer->set(EE, EET, $tentative_events, $self->recorded_date),
+        $self->chronicle_writer->set(EE, EE,  $existing_events,  $self->recorded_date));
+}
```
basiliscos: See my previous comment about that kind of return.

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+    #now combine received data with $events
+    for my $doc (@{$documents}) {
+        #combine $doc->{events} with current $events
+        my $doc_events = $doc->{events};
+        for my $doc_event (@{$doc_events}) {
+            $doc_event->{id} =
+                _generate_id(Date::Utility->new($doc_event->{release_date})->truncate_to_day()->epoch
+                    . $doc_event->{event_name}
+                    . $doc_event->{symbol}
+                    . $doc_event->{impact})
```
basiliscos: It would be better unwrap the code with variable:

```perl 

my $unique_data = join($doc_event->{release_date})->truncate_to_day()->epoch, 
  $doc_event->{event_name}, 
  $doc_event->{symbol},
  $doc_event->{impact}
);
$doc_event->{id} = _generate_id($unique_data);
```

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+                unless defined $doc_event->{id};
```
basiliscos: `$doc_event->{id}` cannot be undefined, doesn't it?

lib/Quant/Framework/EconomicEventCalendar.pm
```diff
@@ -0,0 +1,284 @@
...
+
+            # historical event's release date could still be string.
+            my $doc_release_epoch = Date::Utility->new($doc_event->{release_date})->epoch;
+            $all_events{$doc_event->{id}} = $doc_event if ($doc_release_epoch >= $from and $doc_release_epoch <= $to);
+        }
+    }
+
+    my @result = values %all_events;
+    return \@result;
+}
```
basiliscos: It should be mentioned in method comments, that events might follow in non-historical order (random order). If this is not desirable, they should be sorted by `release_date` before return.

mm-binary:
Thanks @basiliscos . As this PR aims to move the module to Quant::Framework, I will note these comments to be applied to this module (and other similar modules of Quant::Framework) after movement.

PR #15, opened 2016-04-27
Moving Asset and Currency to QF.

shahrizal-shaari:


shahrizal-shaari:
Will recreate a new PR.

PR #16, opened 2016-05-03
Moving Asset and Currency to QF

shahrizal-shaari:


t/currency.t
```diff
@@ -0,0 +1,80 @@
...
+		symbol => 'USD',
+		for_date => Date::Utility->new('2014-10-10'),
+                chronicle_reader    => $chronicle_r,
+                chronicle_writer    => $chronicle_w,
+		});
+
+        my $tiy   = 365.0/365.0;
+        my $rates = $usd->rate_for($tiy);
+
+        print "rates: $rates\n";
```
mm-binary: Do you need these print statement in the unit test?

PR #20, opened 2016-05-11
yngshan/include_otc_submarket

yngshan:
- remove is_otc from exchange


basiliscos:
I think `Changelog` should be maintained :)

jy-binary:
pls check conflicts

jy-binary:
conflicts now

PR #21, opened 2016-05-12
Chylli/copy on write2

chylli:


lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -201,13 +204,18 @@ has trading_days_list => (
     lazy_build => 1,
 );
 
+my $trading_days_aliases;
+
+BEGIN {
+    $trading_days_aliases = YAML::XS::LoadFile(File::ShareDir::dist_file('Quant-Framework', 'exchanges_trading_days_aliases.yml'));
+}
+
```
basiliscos: Could, you, please, move the `BEGIN` block in the top of the file, to ease reader catch, what's going on.

chylli: I think its better to keep  it here because when user read the function `_build_trading_days_list` he can easily catch what's the content of variable `traiding_days_aliases`

PR #25, opened 2016-05-16
moving some helper methods to Q::F

mm-binary:


lib/Quant/Framework/Utils/UnderlyingConfig.pm
```diff
@@ -0,0 +1,244 @@
...
+
+Locale code, used to generate some descriptions in TradingCalendar
+
+=cut
+
+has locale => (
+    is      => 'ro',
+    isa     => 'Str',
+);
+
```
basiliscos: Is it possible not to have locale in Q::F at all? I think it is bad idea, because, you'll need to localize strings here too, and we'll have "non-centralized" i18n strings.

I understand, that you need localize errors. So, I propose to have localization closure instead of locale itself, i.e. 

```perl
has localization_fn => (required => 1);


...
if ($error_condition) {
   my $message = $self->localization_fn("quants.framework.error_condition", "argument1" , "argument2" ,..);
  die ($message);
}

```

basiliscos:
The code itself looks good.

lib/Quant/Framework/Utils/UnderlyingConfig.pm
mm-binary: @basiliscos That's right but locale is needed to create instance of TradingCalendar which is already released. I have a plan to remove this dependency from TradingCalendar in my task-list.

shahrizal-shaari:
@mm-binary Just want to confirm, there is no need to add unit test for the newly added classes (Builder and UnderlyingConfig) ?
The rest looks good.

mm-binary:
@shahrizal-shaari yes. UnderlyingConfig does not have code and for Builder, the methods are being tested in other repositories (already existing unit tests). I may later move some of them to Q::F when vol-surface is moved to Q::F

lib/Quant/Framework/Utils/Builder.pm
```diff
@@ -0,0 +1,307 @@
...
+=head2 dividend_rate_for
+
+Get the dividend rate for this underlying over a given time period (expressed in timeinyears.)
+
+=cut
+
+sub dividend_rate_for {
+    my ($self, $tiy) = @_;
+
+    die 'Attempting to get interest rate on an undefined currency for ' . $self->underlying_config->symbol
```
jy-binary: This message "Attempting to get interest rate on an undefined currency" makes no sense. Dividends are for stocks, not currencies. Why does it say "currency" in the message?

lib/Quant/Framework/Utils/Builder.pm
```diff
@@ -0,0 +1,307 @@
...
+
+    my %zero_rate = (
+        smart_fx  => 1,
+        smart_opi => 1,
+    );
+
+    my $rate;
+
+    if ($self->underlying_config->market_name eq 'volidx') {
+        my $div = $self->build_dividend();
```
jy-binary: i don't understand the code, but logically this is making no sense - why are we building dividends for Volatility Indices?

lib/Quant/Framework/Utils/UnderlyingConfig.pm
```diff
@@ -0,0 +1,244 @@
...
+    required=> 1,
+);
+
+=head2 system_symbol
+
+Internal symbol name
+
+=cut
+
+has system_symbol => (
```
jy-binary: how is the "internal symbol" different to the "symbol" ?

Remember this is an open-source module. It needs to be clear, and for the public domain. This PR introduces a lot of mess, it seems to me.

lib/Quant/Framework/Utils/UnderlyingConfig.pm
```diff
@@ -0,0 +1,244 @@
...
+    isa     => 'Str',
+);
+
+=head2 extra_vol_diff_by_delta
+
+Extra volatility difference
+
+=cut
+
+has extra_vol_diff_by_delta => (
```
jy-binary: what on earth is this? Imagine you are a member of the Perl community trying to make sense of this CPAN module. Would you understand what "extra volatility difference" means?

lib/Quant/Framework/Utils/UnderlyingConfig.pm
```diff
@@ -0,0 +1,244 @@
...
+    is      => 'ro',
+);
+
+=head2 market_convention
+
+Market convention settings
```
jy-binary: where is the explanation for this? What market conventions are we talking about?

mm-binary: Added documentation.

lib/Quant/Framework/Utils/UnderlyingConfig.pm
mm-binary: Documentation enhanced.

lib/Quant/Framework/Utils/UnderlyingConfig.pm
mm-binary: Documentation enhanced.

lib/Quant/Framework/Utils/Builder.pm
mm-binary: Fixed.

lib/Quant/Framework/Utils/Builder.pm
mm-binary: This part is moved back to Underlying module. I think we need a default dividend of zero for R_xx symbols and also a default non-zero value to simulate drift in RDBULL/RDBEAR symbols.

jy-binary:
Looks good to me now. Thx.

lib/Quant/Framework/Utils/UnderlyingConfig.pm
```diff
@@ -0,0 +1,284 @@
...
+ - indices
+ - commodities
+
+=cut
+
+has market_name => (
+    is      => 'ro',
+    isa     => 'Str',
+);
+
```
shahrizal-shaari: @mm-binary @basiliscos , I am just curious, in this kind of case, is there a simple way in perl we can enforce that only those three values(forex,indices,commodities) are valid ?

basiliscos: @shahrizal-shaari  You can write your validation function like (for Moo):

```perl5
 isa => sub {
   my ($self, $value) = @_;
   die("wrong market name: $value") unless $value =~ /(forex)|(indices)|(commodies)/;
 }
```

basiliscos: Something like that should be available for Moose too (may be exactly as I've shown for Moo)

shahrizal-shaari: Thanks @basiliscos .
@basiliscos @mm-binary , so for this one, is it better to enforce or just leave it as it is ?

basiliscos: @shahrizal-shaari 

As this is open-source module, let it be used by people for some other markets. IMHO.

PR #28, opened 2016-05-24
Ivan/refactor qf

basiliscos:
It is not ready yet


lib/Quant/Framework/Document.pm
```diff
@@ -0,0 +1,95 @@
...
+    required => 1,
+);
+
+has for_date => (
+    is      => 'ro',
+    isa     => sub {
+      die("Quant::Framework::Document::for_date should be Date::Utility")
+        unless ref($_[0]) eq 'Date::Utility';
+    },
+    required => 1,
```
mm-binary: This should not be marked as required. Because the convention is that when `for_date` is undef, the caller is looking for the latest version of data.

lib/Quant/Framework/StorageAccessor.pm
```diff
@@ -0,0 +1,30 @@
...
+
+has chronicle_reader => (
+    is       => 'ro',
+);
+
+has chronicle_writer => (
+    is       => 'ro',
+);
+
+1;
```
mm-binary: My `pre_vol_surface` just released which has a new module named `Builder`. This module also has chronicle_reader and writer. Can we merge these two or maybe inherit Builder from `StorageAccessor`?

lib/Quant/Framework/CorporateAction.pm
```diff
@@ -1,228 +1,141 @@
...
+    my $document = Quant::Framework::Document->new(
+        storage_accessor => $storage_accessor,
+        for_date         => $for_date,
+        symbol           => $symbol,
+        data             => { actions => {} },
+    );
+
+    return __PACKAGE__->new(
+      document => $document,
+    );
```
mm-binary: This creates a document with given data but does not save it. right? If so please add appropriate pod docs for the method.

lib/Quant/Framework/CorporateAction.pm
```diff
@@ -1,228 +1,141 @@
...
-);
-
-sub _build_actions {
-    my $self = shift;
-
-    my $document = $self->document;
-
-    return $document->{actions} if defined $document;
-    return {};
+    $self->document->save('corporate_actions');
```
mm-binary: `document` already knows it's namespace, so why sending it again here?

lib/Quant/Framework/Document.pm
basiliscos: @mm-binary That convention is used for `load` method, see

https://github.com/basiliscos/perl-Quant-Framework/blob/ivan/refactor-qf/lib/Quant/Framework/Document.pm#L70
https://github.com/basiliscos/perl-Quant-Framework/blob/ivan/refactor-qf/lib/Quant/Framework/CorporateAction.pm#L82

but the document itself always has the timestamp of it's creation, that's why this field is mandatory for document

lib/Quant/Framework/StorageAccessor.pm
basiliscos: Yes, of course. Although, the use (or embed or composition) is generally more preferable (flexible) over inheritance. But surely, we'll avoid code duplication with those two classes.

lib/Quant/Framework/Document.pm
```diff
@@ -0,0 +1,95 @@
...
+
+=cut
+
+
+has storage_accessor => (
+    is => 'ro',
+    required => 1,
+);
+
+has for_date => (
```
mm-binary: Its better to use `recorded_date` for this field. Because we already have for_date in Underlying which has a different meaning.

lib/Quant/Framework/CorporateAction.pm
basiliscos: Very good point! I'll add namespace to the document

lib/Quant/Framework/CorporateAction.pm
basiliscos: done

lib/Quant/Framework/Document.pm
```diff
@@ -0,0 +1,150 @@
...
+
+Quant::Framework::Document -
+
+=head1 DESCRIPTION
+
+Internal representation of persistend data. Do not B<create> the class directly outside
+of Quant::Framework, although the usage of public fields outside of Quant::Framework
+is allowed.
+
+The class is responsible for loading and stoing data via Data::Chronicle. The
```
shahrizal-shaari: Typo for storing.

share/test_data.yml
```diff
@@ -1,11 +1,4 @@
 ---
-corporate_action:
-  class_name: 'Quant::Framework::CorporateAction'
-  data:
-    symbol: 'USAAPL'
-    actions:
-      80004829: {type: 'DVD_CASH', flag: 'N', description: 'Test Corporate Action', modifier: 'divide', value: 1.234}
-
```
shahrizal-shaari: @mm-binary  Just want to confirm , Is it ok to remove this from test_data.yml ?

lib/Quant/Framework/Document.pm
basiliscos: Fixed, thank you

share/test_data.yml
basiliscos: At the end of refactoring, I expect whole `test_data.yml` & Test.pm will be removed :)

Because this complicates **unit-** testing a lot, and makes it hard to track down, where data comes from

mm-binary: @basiliscos I don't think replacing this yml file with in-the-code data generation is a good idea. Some items in this yml file are quite large and used multiple times in multiple test-cases. If you paste it's data in all those places, it would be a mess.

yngshan: @basiliscos , I think we need a yml because often people might not aware of the format of each doc. May be we should not make it initial.

basiliscos: @mm-binary I have analyzed the usage of Corporate Action, and the file is use-less. That's why I have deleted it. In the most of cases it makes details obscure, and this is very bad. Example:

[1] https://github.com/regentmarkets/bom-market/blob/master/t/BOM/Market/01_corporate_action.t

[2] https://github.com/basiliscos/bom-market/blob/ivan/refactor-qf-corporate-actions/t/BOM/Market/01_corporate_action.t

From [1] it is very hard to figure out, why `USSAPL` is inserted and with what values. 

So, let's do that on-per value basis. The sample yaml isn't large, so, if we replace some parts of it with hard-coded values in some tests, that might become more clear, actually.

basiliscos: @yngshan Good point. Let's keep yaml for reference. At the end of refactoring, it might be moved from shared  to tests only. I'll roll-back the lines removal.

lib/Quant/Framework/StorageAccessor.pm
```diff
@@ -0,0 +1,30 @@
+package Quant::Framework::StorageAccessor;
+
+=head1 NAME
+
+Quant::Framework::StorageAccessor - This class incorporates chronicle accessors
+
+=head1 DESCRIPTION
+
+ my $storage_accessor = Quant::Framework::StorageAccessor->new(
+  chronicle_reader => ...,
```
shahrizal-shaari: Is it ok to use ... here instead of some more real example ?

lib/Quant/Framework/Document.pm
```diff
@@ -0,0 +1,150 @@
...
+has symbol => (
+    is       => 'ro',
+    required => 1,
+);
+
+=head1 SUBROUTINES
+
+=head2 load($storage_accessor, $namespace, $symbol, $for_date)
+
+ my $recent_document = Quant::Framework::Document::load($storage_accessor, 'corporate_actions', "USAAPL");
```
shahrizal-shaari: I think USAAPL is a binary.com symbol. May be we should use something more generic like AAPL.

lib/Quant/Framework/StorageAccessor.pm
basiliscos: The creation of Chronicle objects is out of the scope for this module.

lib/Quant/Framework/Document.pm
basiliscos: Good point, but should be done later. It is not the purpose of the refactoring.

lib/Quant/Framework/Document.pm
```diff
@@ -0,0 +1,150 @@
...
+
+Internal representation of persistend data. Do not B<create> the class directly outside
+of Quant::Framework, although the usage of public fields outside of Quant::Framework
+is allowed.
+
+The class is responsible for loading and storing data via Data::Chronicle. The
+data itself is a hash, which content is provided by users of the class (i.e. by
+CorporateActions).
+
+ # create new (trancient / not-yet-persisted) Document
```
mm-binary: Typo

lib/Quant/Framework/Document.pm
```diff
@@ -0,0 +1,150 @@
...
+
+=cut
+
+sub load {
+    my ($storage_accessor, $namespace, $symbol, $for_date) = @_;
+
+    my $data = $storage_accessor->chronicle_reader->get($namespace, $symbol)
+      or return;
+
+    if ($for_date && $for_date->datetime_iso8601 lt $data->{date}) {
```
mm-binary: Here you are assuming `{date}` is iso_8601 but its not the case for all chronicle docs.

lib/Quant/Framework/Document.pm
```diff
@@ -0,0 +1,150 @@
...
+  Date::Utility->new, # optional
+ )
+
+=cut
+
+=head1 ATTRIBUTES
+
+=head2 storage_accessor
+
+Chonicle assessor
```
evostrov: typo

lib/Quant/Framework/Document.pm
basiliscos: Can you give an example? Actually the best way would be to have epoch at all.

mm-binary: For economic events we store epoch. But problem is, there are lots of historical data in DB which have iso_8601. So I think we should support both types (at least when reading).

lib/Quant/Framework/CorporateAction.pm
```diff
@@ -1,228 +1,220 @@
...
+        if ($action->{flag} eq 'D' and $original->{actions}->{$action_id}) {
             $cancelled{$action_id} = $action;
+            delete $merged_actions{$action_id};
         }
     }
 
-    return \%cancelled;
+    $data->{actions} = \%merged_actions;
+
+    my $new_document = Quant::Framework::Document->new(
```
mm-binary: You can replace this and next 5-6 lines with a call to `Quant::Framework::CorporateAction::create`.

lib/Quant/Framework/CorporateAction.pm
```diff
@@ -1,228 +1,220 @@
...
-sub _build_document {
-    my $self = shift;
+Might return undef, if no actions exists in Chronicle
 
-    my $document = $self->chronicle_reader->get('corporate_actions', $self->symbol);
+=cut
 
-    if ($self->for_date and $self->for_date->datetime_iso8601 lt $document->{date}) {
-        $document = $self->chronicle_reader->get_for('corporate_actions', $self->symbol, $self->for_date->epoch);
+sub load {
```
mm-binary: I think `save` and `load` will be pretty similar for all other Q::F modules. Can we extract these to a parent class or Moose Role or something similar and re-use these codes for other modules too?

lib/Quant/Framework/Document.pm
basiliscos: Thanks for details. I think we need to do one-time -script that will update DB and unify all that :)

lib/Quant/Framework/CorporateAction.pm
basiliscos: No, because `create` creates empty document (with no actions).

lib/Quant/Framework/CorporateAction.pm
basiliscos: You are absolutely right, but I'll generalize the `load` and `save` with my next refactoring step (for some other Q::F object).

jy-binary:
pls check conflicts

lib/Quant/Framework/CorporateAction.pm
```diff
@@ -1,228 +1,220 @@
 package Quant::Framework::CorporateAction;
 
-use Data::Chronicle::Reader;
-use Data::Chronicle::Writer;
+use strict;
+use warnings;
+
+use Date::Utility;
+use Quant::Framework::Document;
+use Moo;
```
tfoertsch: Just out of curiosity, does it make sense to use `Moo` here and `Moose` in most of the other modules? I read this <https://metacpan.org/pod/Moo#MOO-AND-MOOSE>. `Moo` is all about less setup cost. I have no idea what *If Moo detects Moose being loaded, it will automatically register metaclasses for your Moo and Moo::Role packages* really means. But if it means that if `Moo` detects `Moose` it silently turns into `Moose`, then there is no point in using `Moo`.

mm-binary:
@basiliscos Please increase version number.

lib/Quant/Framework/CorporateAction.pm
basiliscos: `Moose` eats additionally +13 Mb per process, comparing to `Moo`, it has 100+ dependencies, and slower start up. That's why I hope sometime get rid of `Moose`.

The compatibility you quote, I understand something like that:

```perl
sub moo_compile_class {
  my $new_class_or_role = shift;
  ...;
  my $moose_registry = eval { "Moose::Class::Registry::get()" };
  if ( $moose_registry) {
   $moose_registry->register($new_class_or_role);
 }
 ..
}

```

I.e. I don't think it adds any weight to memory or startup time.

PR #29, opened 2016-05-25
yngshan/corporate_action_fix

yngshan:


lib/Quant/Framework/Utils/Builder.pm
```diff
@@ -246,7 +248,7 @@ sub get_discrete_dividend_for_period {
         }
     }
 
-    return \%valid_dividends;
+    return ($dividend_recorded_date, \%valid_dividends);
```
basiliscos: As this is public method, please, update the return description for it.

shahrizal-shaari: And I think, don't forget to update the file called "Changes".

mm-binary:
LGTM

PR #35, opened 2016-06-15
yngshan/add_exchange_for_metal

yngshan:


basiliscos:
conflicts

lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -797,7 +797,7 @@ sub standard_closing_on {
 
     $when = $self->trading_date_for($when);
 
-    return $self->closes_early_on($when) if ($self->symbol eq 'FOREX' and $when->day_of_week == 5);
+    return $self->closes_early_on($when) if (($self->symbol eq 'FOREX' or $self->symbol eq 'METAL') and $when->day_of_week == 5);
```
junbon: Are there any similarity between METAL and FOREX? Instead of checking for FOREX and METAL everywhere maybe we can do $self->similar_behavior as a check?

share/test_underlying_config.yml
```diff
@@ -73,7 +73,7 @@ frxXAUUSD:
   quanto_only : 0
   rate_to_imply_from : 'USD'
   volatility_surface_type : 'delta'
```
junbon: Just checking, metal has delta surface?

yngshan: Yes, metals will be those gold, silver , palladium and platinum

lib/Quant/Framework/TradingCalendar.pm
yngshan: Not really except trading timezone and trading hour which do not serve as a unique similarity

PR #36, opened 2016-06-16
Refactor Holiday object

basiliscos:


lib/Quant/Framework/CorporateAction.pm
```diff
@@ -4,9 +4,10 @@ use strict;
 use warnings;
 
 use Date::Utility;
-use Quant::Framework::Document;
 use Moo;
 
+with('Quant::Framework::Document');
```
tfoertsch: I am fine with that change. It does change the interface a bit. So, I think a major version bump is appropriate. Also, please point out that change in the documentation.

jy-binary:
got conflicts now

lib/Quant/Framework/Holiday.pm
```diff
@@ -1,105 +1,138 @@
...
+            while (my ($description, $new_symbols) = each %$new_holiday) {
+                my @merged_symbols = uniq(@${$persisted_event->{$description}}, @$new_symbols);
+                $persisted_event->{$description} = \@merged_symbols;
+            }
         }
     }
 
-    return $self->chronicle_writer->set('holidays', 'holidays', \%{$cached_holidays}, $self->recorded_date);
+    my $new_obj = __PACKAGE__->new(
+        data             => {calendar => \%actual_events},
```
yngshan: This should be %actual_events or %persisted_event?

lib/Quant/Framework/Document.pm
```diff
@@ -4,41 +4,23 @@ use strict;
 use warnings;
 use Date::Utility;
 
-use Moo;
+use Moo::Role;
 
 =head1 NAME
 
-Quant::Framework::Document - Binds data with Chronicle
```
mm-binary: Typo

lib/Quant/Framework/Document.pm
```diff
@@ -96,53 +72,108 @@ has symbol => (
...
+sub load_default {
+    my ($package, $storage_accessor, $symbol, $for_date) = @_;
+
+    my $namespace = $package->namespace;
 
     my $data = $storage_accessor->chronicle_reader->get($namespace, $symbol)
-      or return;
+        or return;
 
     if ($for_date && $for_date->datetime_iso8601 lt $data->{date}) {
```
mm-binary: It's preferred to compare epochs rather than strings.

lib/Quant/Framework/Document.pm
```diff
@@ -96,53 +72,108 @@ has symbol => (
...
+=head2 default_section()
+
+The section in data hash (string), where object-specific data will be stored.
+For example for Holidays, it can be 'calendar'
+
+=cut
+
+requires 'namespace';
+
+requires 'default_section';
```
mm-binary: This name is a little misleading (When something is labeled `default` means it can be overridden. But these data are not supposed to be overridden anywhere). Maybe `custom_data` or `internal_document`?

lib/Quant/Framework/Document.pm
```diff
@@ -96,53 +72,108 @@ has symbol => (
...
+sub create_default {
+    my ($package, $storage_accessor, $symbol, $for_date) = @_;
+    my $obj = $package->new(
+        storage_accessor => $storage_accessor,
+        recorded_date    => $for_date,
+        symbol           => $symbol,
+        data             => {$package->default_section => {}},
+    );
+}
+*create = \&create_default;
```
mm-binary: Why do we need to have these additional methods here? Can't we just have default implementations for `create` and other similar methods, and let child classes override those method if they need?

lib/Quant/Framework/Document.pm
```diff
@@ -96,53 +72,108 @@ has symbol => (
...
- $document->save;
-
-Stores (persists) the document in Chronicle database.
+Stores (persists) the object in Chronicle database.
 
 =cut
 
 sub save {
     my $self = shift;
     # the most probably this is redundant, anc can be removed in future
```
mm-binary: Typo

lib/Quant/Framework/Document.pm
mm-binary: Also, I think this won't be useful when you move to other QF modules. Take a look at vol-surface. It has more than one of so called `default_sections`. Why not drop this and let the child class add as many elements to `$self->data` as it wants?

lib/Quant/Framework/Holiday.pm
```diff
@@ -1,105 +1,138 @@
...
-    my $calendar =
-        ($for_date) ? $reader->get_for('holidays', 'holidays', $for_date) : 
-        $reader->get('holidays', 'holidays');
     my %holidays;
-    foreach my $date (keys %$calendar) {
-        foreach my $holiday_desc (keys %{$calendar->{$date}}) {
-            $holidays{$date} = $holiday_desc if (first { $symbol eq $_ } @{$calendar->{$date}{$holiday_desc}});
+    while (my ($epoch, $holiday) = each %$data) {
+        while (my ($description, $symbols) = each %$holiday) {
+            $holidays{$epoch} = $description if (first { $symbol eq $_ } @$symbols);
```
mm-binary: I know this is outside scope of this change but this is a highly inefficient code (O(n^3) execution time). Is it possible to simplify this section?

lib/Quant/Framework/Document.pm
```diff
@@ -96,53 +72,108 @@ has symbol => (
...
+  Quant::Framework::CorporateAction->load($storage_accessor, 'QWER')
+
+Please note, it should be invoked with package name (the module, to which the
+role is applied to), otherwise it will not work.
 
 =cut
 
-sub load {
-    my ($storage_accessor, $namespace, $symbol, $for_date) = @_;
+sub create_default {
```
mm-binary: Can't we just eliminate `create` method and let default constructor of the module do this job?
In this case `Quant::Framework::Holiday->new` will do whatever `create` is supposed to do (create a default empty instance).

lib/Quant/Framework/Holiday.pm
```diff
@@ -1,105 +1,138 @@
...
+    my ($package, $storage_accessor, $for_date) = @_;
+    $package->create_default($storage_accessor, namespace, $for_date);
+}
+
+sub load {
+    my ($package, $storage_accessor, $for_date) = @_;
+    $package->load_default($storage_accessor, namespace, $for_date);
+}
+
+sub update {
```
mm-binary: This is not a big deal but maybe you can rename this method. Seems it is doing something different.

lib/Quant/Framework/Document.pm
```diff
@@ -57,37 +39,31 @@ The date of document creation (C<Date::Utility>)
...
 The domain-specific name of document; e.g. "USAAPL" for corporate actions
 
+=head2 namespace
+
 =cut
 
 has storage_accessor => (
-    is => 'ro',
+    is       => 'ro',
     required => 1,
```
mm-binary: You can check to make sure only one of `chronicle_reader` or `chronicle_writer` is passed.

lib/Quant/Framework/Document.pm
```diff
@@ -96,53 +72,108 @@ has symbol => (
...
+
+Please note, it should be invoked with package name (the module, to which the
+role is applied to), otherwise it will not work.
 
 =cut
 
-sub load {
-    my ($storage_accessor, $namespace, $symbol, $for_date) = @_;
+sub create_default {
+    my ($package, $storage_accessor, $symbol, $for_date) = @_;
```
mm-binary: Why does create method need a `for_date`?

lib/Quant/Framework/Document.pm
basiliscos: because child method might still need to invoke the default implementation. See 

https://github.com/basiliscos/perl-Quant-Framework/blob/ivan/refactor-holydays/lib/Quant/Framework/Holiday.pm#L68

lib/Quant/Framework/Document.pm
basiliscos: fixed

lib/Quant/Framework/Document.pm
basiliscos: Well, yes, I agree with you here. 

This method is required for default construction of object instance in Document::create, see:

https://github.com/basiliscos/perl-Quant-Framework/blob/ivan/refactor-holydays/lib/Quant/Framework/Document.pm#L140

So, let's rename it to `data_initializer` which will return hash-with defaults. OK?

lib/Quant/Framework/Document.pm
basiliscos: done

lib/Quant/Framework/Holiday.pm
basiliscos: `%actual_events` = `%persisted_events` + `%new_events`, plus filtered events, which are older than `$recoreded_date`.

The line 92 and 98 operates on `persisted_event` (taken from actual, which is clone of persisted in line 84). The logic is following if the persisted / old event isn't present for the `$new_date`., then we just append new event. Otherwise, we just append new symbols from `$new_event` to the persisted one.

lib/Quant/Framework/Holiday.pm
basiliscos: I doubt that it is possible, because data structure has depth of 3, i.e. `{ epoch => { "event-name" => [ qw/USD JPY ] }}` and the method requires filtering by the most deep element.

lib/Quant/Framework/Document.pm
basiliscos: Because `recorded_date` is required argument.

lib/Quant/Framework/Document.pm
basiliscos: How would you **load** object if only `chronicle_writer` is passed?

lib/Quant/Framework/Holiday.pm
basiliscos: Which name do you suggest? I find the `update` fits well for this object as well as for all other objects, because it accepts some other data-hash, merges it with internal data, and returns new (unsaved) object.

lib/Quant/Framework/Document.pm
mm-binary: This is ok.

lib/Quant/Framework/Document.pm
mm-binary: Can't we do it like this?
`$self->SUPER::create(...);`

lib/Quant/Framework/Holiday.pm
mm-binary: Maybe `merge_with`. Because it better describes what the methods is doing. But as I said it's not a big deal.

lib/Quant/Framework/Document.pm
mm-binary: If someone passes only the writer, the it doesn't want to `load` data. It just want to save some fresh and newly generated data.

lib/Quant/Framework/Document.pm
mm-binary: Ok. So make it optional. When nothing is passed it will assume current date.

lib/Quant/Framework/Document.pm
mm-binary: Here we have 3 methods (create, load and save) but we define 6 methods to do the job. And this makes things complicated.

mm-binary: Maybe we can use this: http://search.cpan.org/~tobyink/Class-Method-ModifiersX-Override-0.002/lib/Class/Method/ModifiersX/Override.pm
Or simply use around in child classes.

basiliscos: We can't invoke it via `$self->SUPER::create(...)` because this is *role*, not parent. The method is copied on comple-time from role to class, unless class defines it's own implementation.

basiliscos: > Here we have 3 methods (create, load and save) but we define 6 methods to do the job. And this makes things complicated.

Strictly speaking there are 3 methods, 2 additional **aliases** (create/load).

lib/Quant/Framework/Document.pm
basiliscos: No, this is potential data-race. All default object creation, where time is involved, is potential bug. I've seen somewhere in our tests code like:

```perl
my $obj =  ... ->new() # (1)
$obj->save;              #  (2)
````
 It implied date as `time`. So, in most of time (1) and (2) was the same time, but sometimes different. It is typical data race. 

Anyway, conceptually, document must always has a date. If we make it optional, than a user of document, might assume that it can be `undef` (even after load). That's false, and lead user into confusion.

File: lib/Quant/Framework/Document.pm (hunk 9 above)
Comment (basiliscos) [inline lib/Quant/Framework/Document.pm]: Who will initialize the data with defaults(e.g. `calendar => {}` for holidays), then? 

1. The client of `object`? That's not his responsibility, and he can just forgot hat. This is error-prone way.

2. Do not initialize it at all? That means that in every place we use `data` we should have `if`, i.e. `if $data->{calendar}`. That compexifies `Document` and every `Object`, because instead of assuming that data is initialize we have to handle the case when it isn't in **every method** like in old CA:

https://github.com/basiliscos/perl-Quant-Framework/blob/master/lib/Quant/Framework/CorporateAction.pm#L83

https://github.com/basiliscos/perl-Quant-Framework/blob/master/lib/Quant/Framework/CorporateAction.pm#L172

https://github.com/basiliscos/perl-Quant-Framework/blob/master/lib/Quant/Framework/CorporateAction.pm#L91.

3. Do that lazily? It also bad way for the **required** data. Actually lazy-accessors are a bit overused in our system. The usual use-case for lazy-constructors, is to lazily create **optional heavy** properties. They are really cool here then, but in other cases it just spread the object initialization.

File: lib/Quant/Framework/Document.pm (hunk 5 above)
Comment (basiliscos) [inline lib/Quant/Framework/Document.pm]: done

File: lib/Quant/Framework/Document.pm (hunk 11 above)
Comment (basiliscos) [inline lib/Quant/Framework/Document.pm]: I think we never do that, except for the very first object creation. I.e. for holidays:our use case is:

```perl

my $h = Holidays->load(..);
$h_new = $h->update(...);
$h_new->save;

```

While we can do 

```perl

my $h = Holidays->create(..);
$h->save;
```

We'll forget history. Actually, in the old code it **always** loads object before saving, see:

https://github.com/basiliscos/perl-Quant-Framework/blob/master/lib/Quant/Framework/ImpliedRate.pm#L79

https://github.com/basiliscos/perl-Quant-Framework/blob/master/lib/Quant/Framework/Holiday.pm#L55

lib/Quant/Framework/Document.pm
```diff
@@ -86,62 +67,123 @@ has data => (
...
     is       => 'ro',
     required => 1,
 );
 
-=head1 SUBROUTINES
+=head1 REQUIED METHODS
+
+=head2 namespace()
+
+returns namespace (strign) for the object, which the role is applied to,
```
mm-binary: Typo

lib/Quant/Framework/Document.pm
mm-binary: I prefer a hash-ref for passing arguments here because in future the developer might forget what was the order of the arguments.

lib/Quant/Framework/Document.pm
```diff
@@ -86,62 +67,123 @@ has data => (
...
+        recorded_date    => $for_date,
+        symbol           => $symbol,
+        data             => $data,
+    );
+    return $obj;
+}
+*create = \&create_default;
+
+sub load_default {
+    my ($package, $storage_accessor, $symbol, $for_date) = @_;
```
mm-binary: I prefer a hash-ref for passing arguments here because in future the developer might forget what was the order of the arguments.

mm-binary:
Please update README file if needed.

jy-binary:
can you add me as a collaborator so I can see the Update Branch button?

PR #39, opened 2016-06-22
Mm/clean builder

mm-binary:


lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -244,14 +259,15 @@ BEGIN {
...
 
     croak "Exchange symbol must be specified" unless $symbol;
     my $params_ref = clone($exchanges->{$symbol});
-    $params_ref->{symbol}           = $symbol;
-    $params_ref->{for_date}         = $for_date if $for_date;
-    $params_ref->{locale}           = $locale if $locale;
-    $params_ref->{chronicle_reader} = $chronicle_r;
+    $params_ref = {%$params_ref, %$orig_params_ref};
+
+    my $locale = $orig_params_ref->{locale} // 'EN';
```
junbon: Why do we need language in QF?

mm-binary: This is a legacy argument. I have another PR to remove this (mm/remove_locale)

lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -1280,27 +1296,113 @@ sub trading_period {
...
+}
+
+=head2 extended_weight_on
+
+Returns the weight for a given day (given as a Date::Utility object).
+Returns our closed weight for days when the market is closed.
+
+=cut
+
+sub extended_weight_on {
```
junbon: I am confused with ```simple_weight_on``` and ```extended_weight_on``` functions. Can we just have one called ```weight_on```? This function should only be used in volatility adjustment if I am not mistaken.

lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -1280,27 +1296,113 @@ sub trading_period {
...
+
+    return unless $self->underlying_config->asset_symbol;
+    my $type = $self->underlying_config->asset_class;
+
+    my $which = $type eq 'currency' ? 'Quant::Framework::Currency' : 'Quant::Framework::Asset';
+
+    return $which->new({
+        symbol           => $self->underlying_config->asset_symbol,
+        for_date         => $self->for_date,
+        chronicle_reader => $self->chronicle_reader,
```
junbon: I am still a little bit confused as to when we need ```chronicle_reader``` and ```chronicle_writer```. In this case, why writer is not required? I still the ```Quant::Framework::*``` interface needs to be more standardized.

mm-binary: Yes. But TradingCalendar does not need and is not supposed to use a `chronicle_writer`. That's why it cannot pass writer to asset it's creating. I think the "consistent interface" will be achieved more when Ivan continues introducing storage_accessor for all remaining QF modues.

lib/Quant/Framework/TradingCalendar.pm
mm-binary: Sure. It's used in `weighted_days_in_period` which is valled for vol interpolation.

mm-binary: The reason I added extended_* is that after moving `weight_on` from Builder to TC, I had two methods both named `weight_on` so I had to differentiate them.

jy-binary:
pls check conflicts

PR #41, opened 2016-06-23
Mm/remove locale

mm-binary:


jy-binary:
tests failing

lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -267,27 +256,22 @@ sub BUILDARGS {
     my $params_ref = clone($exchanges->{$symbol});
     $params_ref = {%$params_ref, %$orig_params_ref};
 
-    my $locale = $orig_params_ref->{locale} // 'EN';
-
     foreach my $key (keys %{$params_ref->{market_times}}) {
         foreach my $trading_segment (keys %{$params_ref->{market_times}->{$key}}) {
             if ($trading_segment eq 'day_of_week_extended_trading_breaks') { next; }
             elsif ($trading_segment ne 'trading_breaks') {
                 $params_ref->{market_times}->{$key}->{$trading_segment} = Time::Duration::Concise::Localize->new(
```
basiliscos: I would prefer to go 1 step further - remove `Time::Duration::Concise::Localize` at all. Because now, you have English-localized interval in Q::F

mm-binary: We are using that module in our code.

basiliscos: It is not used in Q::F (is it?), so should be moved out of here:

```
ack-grep Time::Duration::Concise::Localize
lib/Quant/Framework/TradingCalendar.pm
276:                $params_ref->{market_times}->{$key}->{$trading_segment} = Time::Duration::Concise::Localize->new(
284:                    my $open_int = Time::Duration::Concise::Localize->new(
288:                    my $close_int = Time::Duration::Concise::Localize->new(

```

mm-binary: See `trading_date_for` or `standard_closing_on` in TradingCalendar. There are other examples too. `market_times` is a hash-ref which consists of instances of this module and we are using attributes of `Time::Duration::Concise::Localize` in the code.

basiliscos: Yeah, I see. I think that should be refactored to get rid of `Time::Duration::Concise::Localize`, I don't think we need it in `Q::F`, i.e. specify time intervals in seconds should be enough for this module

PR #42, opened 2016-06-27
Mm/cache weights

mm-binary:


lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -1305,6 +1305,17 @@ Returns the sum of the weights we apply to each day in the requested period.
...
+    state %cache;
+    my $key =
+          $begin->epoch
+        . $end->epoch
+        . $self->underlying_config->quoted_currency_symbol
+        . $self->underlying_config->asset_symbol
+        . $self->symbol
+        . ($self->for_date ? $self->for_date->epoch : 0);
+
+    return $cache{$key} if defined $cache{$key};
```
basiliscos: Please, use attribute `_cache`, instead of `state %cache`. 

The current code will be buggy. I.e. it will calculate `weighted_days` for `$now`, but if the object has been refreshed because of new holidays, the `$key` will be the same, hence, will return previous value. 

Having `_cache` attribute leads to object rebuild every `30s`, hence, no problem arises.

jy-binary:
can you add me as a collaborator so I can see the Update Branch button?

lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -1305,6 +1286,20 @@ Returns the sum of the weights we apply to each day in the requested period.
...
+    my $key =
+          $begin->epoch
+        . $end->epoch
+        . $self->underlying_config->quoted_currency_symbol
+        . $self->underlying_config->asset_symbol
+        . $self->symbol
+        . ($self->for_date ? $self->for_date->epoch : 0);
+
+    #empty cache on 5-minute slices so upon updating related data, the cache will be refreshed
+    %cache = () if (time % 300 == 0);
```
evostrov: Should you empty cache every 5-minutes slices?
Do you sure that `time % 300 == 0` code will true every 5-minutes slices?

mm-binary: Good point. I updated the algorithm to use "age" instead of current time.

PR #45, opened 2016-07-01
junbon/volsurface_refactoring

junbon:


lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -1330,6 +1330,19 @@ Returns our closed weight for days when the market is closed.
...
+
+    # clears cache after 5 minutes
+    if (time - $cache_init_time > 300) {
+        $cache_init_time = time;
+        %cache           = ();
+    }
+
+    my $epoch = $date->truncate_to_day->epoch;
+
+    return $cache{$self->symbol}{$epoch} if exists $cache{$self->symbol} and exists $cache{$self->symbol}{$epoch};
```
basiliscos: it should be enough to check `if exists  $cache{$self->symbol}{$epoch}`

lib/Quant/Framework/VolSurface.pm
```diff
@@ -48,11 +47,43 @@ has chronicle_writer => (
...
+has document => (
+    is         => 'rw',
+    lazy_build => 1,
+);
+
+sub _build_document {
+    my $self = shift;
+
+    return $self->chronicle_reader->get_for('volatility_surfaces', $self->symbol, $self->for_date->epoch) // {} if $self->for_date;
+    return $self->chronicle_reader->get('volatility_surfaces', $self->symbol) // {};
```
basiliscos: AFAIK, general logic is:
1. fetch object from redis
2. if it not fresh enough for `for_date`, then fetch object DB.

In your case you directly try to fetch it from DB if `for_date` is defined.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -82,101 +113,77 @@ sub _build_builder {
...
-    isa        => 'Date::Utility',
-    lazy_build => 1,
-);
+sub _build_surface {
+    my $self = shift;
 
-sub _build_recorded_date {
-    my $self          = shift;
-    my $recorded_date = Date::Utility->new($self->document->{date});
+    my $surface_data       = dclone($self->surface_data);
```
basiliscos: What is the need of cloning the object? Why can't you just create a new hash with new calculated data (without duplicates from `$self->surface_data` ) ?

lib/Quant/Framework/VolSurface.pm
```diff
@@ -82,101 +113,77 @@ sub _build_builder {
...
-Stores the represented surface as a data structure (read-only).
+=head2 symbol
 
-Term structures of CALL delta or moneyness to annualized volatilities.
+The symbol of the underlying that this surface is for (e.g. frxUSDJPY)
 
 =cut
 
-has surface => (
+has symbol => (
```
basiliscos: What is the need of caching symbol? Can't you just remove it and access to it as `$self->underlying_config->system_symbol`? Or directly call the function without caching the value in attribute?

lib/Quant/Framework/VolSurface.pm
```diff
@@ -297,48 +273,26 @@ has original_term => (
...
-                and not defined $surface->{$days[$end_index]}->{atm_spread})
-            {
-                $spread_type = 'vol_spread';
-                $surface->{$days[$end_index]}->{vol_spread} = $self->default_vol_spread;
-                push @vol_spreads, $days[$end_index];
-
-            }
+            my $day = $days[$end_index];
+            next if exists $surface->{$day}{vol_spread};
+            $surface->{$day}{vol_spread} = {$self->atm_spread_point => 0.1};
```
basiliscos: What is the `atm_spread_point => 0.1`? Why it appears here at all? Looks as if this is personal JB's backdoor :)

PS. It's better to introduce / use some constant for that.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -608,349 +424,72 @@ Returns the ask/bid spread for smile of this volatility surface
...
-
-    } else {
-
-        $smile_spread = $self->_compute_and_set_smile_spread($day);
-    }
-
-    return $smile_spread;
-
+    return $surface->{$day}{vol_spread} if (exists $surface->{$day} and exists $surface->{$day}{vol_spread});
+    return $self->_compute_and_set_smile_spread($day);
```
basiliscos: `return  $surface->{$day}{vol_spread} // $self->_compute_and_set_smile_spread($day)` ? :)

lib/Quant/Framework/VolSurface/Delta.pm
```diff
@@ -161,307 +149,195 @@ sub get_volatility {
...
-            my $v = $v1 * sqrt($t / $t1);
-            $surface_hashref->{$maturity}->{smile}->{$delta} = $v;
-
-            if (defined $surface1->surface->{$maturity}->{vol_spread}->{$delta}) {
-                $surface_hashref->{$maturity}->{vol_spread}->{$delta} = $surface1->surface->{$maturity}->{vol_spread}->{$delta};
-            }
-
-        }
+    my @available_tenors = sort { $a <=> $b } keys %{$table};
+    my @closest = map { Date::Utility->new($_) } @{find_closest_numbers_around($date->epoch, \@available_tenors, 2)};
```
basiliscos: Why not `my @closest = map { Date::Utility->new($_) }($date->epoch -2 .. $date->epoch -2).

lib/Quant/Framework/VolSurface/Delta.pm
```diff
@@ -161,307 +149,195 @@ sub get_volatility {
...
+    my $weight = $self->get_weight($closest[0], $date);
+    my $weight2 = $weight + $self->get_weight($date, $closest[1]);
 
-        if (my $tenor = $surface1->surface->{$maturity}->{tenor}) {
-            $surface_hashref->{$maturity}->{tenor} = $tenor;
-        }
+    my %variances;
+    foreach my $delta (@{$self->smile_points}) {
+        my $var1 = $table->{$closest[0]->epoch}{$delta};
+        my $var2 = $table->{$closest[1]->epoch}{$delta};
```
basiliscos: closest values 0 and 1 are used only. Why? I expected all 4 should be used? Otherwise, above you can define closes t numbers as `($date->epoch -2, $date->epoch -1)`

lib/Quant/Framework/VolSurface/Moneyness.pm
```diff
@@ -185,31 +151,135 @@ sub get_volatility {
...
+    # a lot more steps to calculate vol for a delta point
+    # on a moneyness surface
+    return $self->_calculate_vol_for_delta($args) if $args->{delta};
+
+    my $moneyness =
+          $args->{strike}
+        ? $args->{strike} / $self->spot_reference * 100
+        : $args->{moneyness};
+
+    die "Sought point must be a number." if not looks_like_number($moneyness);
```
basiliscos: Please make stricter checks. I.e. `$moneyness >0` or something like that. `- infinity` also looks as number :)

lib/Quant/Framework/VolSurface/Moneyness.pm
```diff
@@ -185,31 +151,135 @@ sub get_volatility {
...
+sub get_smile {
+    my ($self, $day) = @_;
+
+    die("day[$day] must be a number.")
+        unless looks_like_number($day);
+
+    my $surface = $self->surface;
+
+    return $surface->{$day}{smile} if (exists $surface->{$day} and exists $surface->{$day}{smile});
+    return $self->_compute_and_set_smile($day);
```
basiliscos: Also can be simplified with `//`

lib/Quant/Framework/VolSurface/Moneyness.pm
```diff
@@ -185,31 +151,135 @@ sub get_volatility {
...
+    my ($self, $T, $T1, $T2) = @_;
+
+    # Implements interpolation over time based on the way Iain M Clark does
+    # it in Foreign Exchange Option Pricing, A Practitioner's Guide, page 70.
+    my $effective_date = $self->effective_date;
+
+    my %dates = (
+        T  => $effective_date->plus_time_interval($T - 1 . 'd23h59m59s'),
+        T1 => $effective_date->plus_time_interval($T1 - 1 . 'd23h59m59s'),
+        T2 => $effective_date->plus_time_interval($T2 - 1 . 'd23h59m59s'),
```
basiliscos: OMG! Why not operate on numbers (`$epoch`)? You create string here from integer, then create new object from string...

lib/Quant/Framework/VolSurface.pm
mm-binary: Why are you returning {} if there is nothing in DB/Redis. Shouldn't the code die in this case?

mm-binary:
Please increase version

lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -1324,6 +1324,19 @@ Returns our closed weight for days when the market is closed.
 sub weight_on {
     my ($self, $date) = @_;
 
+    state %cache;
+    state $cache_init_time = time;
+
+    # clears cache after 5 minutes
+    if (time - $cache_init_time > 300) {
+        $cache_init_time = time;
```
mm-binary: Is this code supposed to be thread-safe?

lib/Quant/Framework/VolSurface.pm
```diff
@@ -82,101 +113,77 @@ sub _build_builder {
...
-    my $self          = shift;
-    my $recorded_date = Date::Utility->new($self->document->{date});
+    my $surface_data       = dclone($self->surface_data);
+    my $expiry_conventions = $self->builder->build_expiry_conventions;
+    my $effective_date     = $self->effective_date;
+
+    # If the smile's day is given as a tenor, we convert
+    # it to a day and add the tenor to the smile:
+    foreach my $maturity (keys %$surface_data) {
+        if ($maturity =~ /^(?:ON|\d{1,2}[WMY])$/) {
```
mm-binary: Better to issue a watninh ig `$maturity` does not confirm to this pattern

lib/Quant/Framework/VolSurface.pm
```diff
@@ -82,101 +113,77 @@ sub _build_builder {
...
+            my $vol_expiry_date = $expiry_conventions->vol_expiry_date({
+                from => $effective_date,
+                term => $maturity
+            });
+            my $day = $vol_expiry_date->days_between($effective_date);
```
mm-binary: So if vol-surfaces's tenor is `1W` and it's effective_data is 'July 1' you will change tenor to 'July 8' right?

lib/Quant/Framework/VolSurface.pm
```diff
@@ -82,101 +113,77 @@ sub _build_builder {
...
+    isa        => 'Date::Utility',
     lazy_build => 1,
 );
 
-sub _build_default_vol_spread {
-    return {shift->atm_spread_point => 0.1};
+sub _build_recorded_date {
+    my $self = shift;
+
+    return Date::Utility->new($self->document->{date});
```
mm-binary: Is this `date` the date on which we saved the vol-surface on our side or the original creation date of the surface?

lib/Quant/Framework/VolSurface.pm
junbon: @basiliscos I think the general logic might have been misunderstood. The usage of ```for_date``` must only be done for back pricing. So if ```for_date``` is present, we should fetch it directly from the DB. For live pricing in production, we should always fetch it from redis. Or am I the one misunderstanding this? @mm-binary what do you think?

lib/Quant/Framework/TradingCalendar.pm
junbon: Do you see a problem?

lib/Quant/Framework/TradingCalendar.pm
junbon: noted

lib/Quant/Framework/VolSurface.pm
junbon: tenor will remain as '1W', we are converting tenors to number of days. So '1W' will be replaced by '7'.

lib/Quant/Framework/VolSurface.pm
junbon: I don't want ```$self->surface_data``` to be modified. Am I misusing ```dclone```?

lib/Quant/Framework/VolSurface.pm
junbon: noted

lib/Quant/Framework/VolSurface.pm
```diff
@@ -1469,24 +712,100 @@ has validation_error => (
...
+=head2 save
+Saves current surface using given chronicle writer.
+=cut
+
+sub save {
+    my $self = shift;
+
+    #if chronicle does not have this document, first create it because in document_content we will need it
+    if (not defined $self->chronicle_reader->get('volatility_surfaces', $self->symbol)) {
+        #Due to some strange coding of retrieval for recorded_date, there MUST be an existing document (even empty)
```
mm-binary: Do we still need to write empty vol-surface in this case?

lib/Quant/Framework/VolSurface.pm
junbon: All market_data has a symbol as attribute. Do we want to do it for everything or keep it in this manner? If we want to change it, I propose we do it in another branch.

lib/Quant/Framework/VolSurface.pm
junbon: This ```$self->document->{date}``` is the original creation date of the surface.

lib/Quant/Framework/VolSurface.pm
junbon: I like my version better 👯

lib/Quant/Framework/VolSurface/Delta.pm
junbon: No. We are trying to find the closest two tenors on the volatility surface and not just any two dates.

lib/Quant/Framework/VolSurface/Delta.pm
junbon: ```@closest``` contains the two dates that we want to use to calculate the new variance.

lib/Quant/Framework/VolSurface/Moneyness.pm
junbon: The check for ```$moneyness``` cannot be negative is in the next line.

lib/Quant/Framework/VolSurface/Delta.pm
```diff
@@ -161,307 +149,195 @@ sub get_volatility {
...
         :                             $self->_convert_moneyness_to_delta($args);
 
-    return $self->SUPER::get_volatility({
-        sought_point => $sought_point,
-        days         => $args->{days},
-    });
-}
-
-=head2 interpolate
+    die 'Delta cannot be zero or negative.' if $delta < 0;
```
mm-binary: `if $delta <= 01?

lib/Quant/Framework/VolSurface/Delta.pm
```diff
@@ -161,307 +149,195 @@ sub get_volatility {
...
+    my $smile = $self->get_smile($args->{from}, $args->{to});
 
-sub _convert_moneyness_to_delta {
-    my ($self, $args) = @_;
+    return $smile->{$delta} if $smile->{$delta};
 
-    $args->{strike} = get_strike_for_moneyness({
-        moneyness => $args->{moneyness},
-        spot      => $self->underlying_config->spot
+    return $self->interpolate({
```
mm-binary: Should we cache this output for future calls with same delta?

lib/Quant/Framework/VolSurface/Delta.pm
```diff
@@ -161,307 +149,195 @@ sub get_volatility {
...
+    my $weight2 = $weight + $self->get_weight($date, $closest[1]);
 
-        if (my $tenor = $surface1->surface->{$maturity}->{tenor}) {
-            $surface_hashref->{$maturity}->{tenor} = $tenor;
-        }
+    my %variances;
+    foreach my $delta (@{$self->smile_points}) {
+        my $var1 = $table->{$closest[0]->epoch}{$delta};
+        my $var2 = $table->{$closest[1]->epoch}{$delta};
+        $variances{$delta} = $var1 + ($var2 - $var1) / $weight2 * $weight;
```
mm-binary: Better to use parentheses to stress execution order.

lib/Quant/Framework/VolSurface/Moneyness.pm
junbon: Made the change. Maybe it is better now?

lib/Quant/Framework/VolSurface/Delta.pm
```diff
@@ -161,307 +149,195 @@ sub get_volatility {
...
-
-    my $master_surface = $self->surface;
-    my %surfaces;
-    $surfaces{$self->cutoff->code} = $master_surface;
+sub get_weight {
+    my ($self, $date1, $date2) = @_;
+
+    my ($date1_epoch, $date2_epoch) = ($date1->epoch, $date2->epoch);
+    my $time_diff       = $date2_epoch - $date1_epoch;
+    my $weight_interval = 4 * 3600;
```
mm-binary: Why 4? Add some comments here please.

lib/Quant/Framework/VolSurface/Delta.pm
junbon: This should be fast. I don't think we need to cache it.

lib/Quant/Framework/VolSurface/Delta.pm
junbon: Added comment.

lib/Quant/Framework/VolSurface/Delta.pm
```diff
@@ -18,140 +18,128 @@ use Moose;
...
-    return $self->smile_points;
-}
 
-has atm_spread_point => (
-    is      => 'ro',
-    isa     => 'Num',
-    default => '50',
-);
+    # For backward compatibility
+    my $surface = $self->document->{surface} // $self->document->{surfaces}{'New York 10:00'} // {};
```
mm-binary: Better to die with a good message if `document` is empty.

lib/Quant/Framework/VolSurface.pm
mm-binary: @JB, your logic here seems fine to me.

lib/Quant/Framework/VolSurface.pm
mm-binary: Can't we solve `strange coding of retrieval for recorded_date` issue in this PR?

junbon: LOL, OK I will put more effort :D. I will probably do it in another branch
so that it is easier to review the changes.

lib/Quant/Framework/VolSurface.pm
basiliscos: @junbon , @mm-binary  Please look at similar code:

https://github.com/binary-com/perl-Quant-Framework/blob/master/lib/Quant/Framework/ImpliedRate.pm#L79

It behaves differently that JB's one.

lib/Quant/Framework/TradingCalendar.pm
basiliscos: Different trading calendar instances will have shared cache?

lib/Quant/Framework/VolSurface.pm
basiliscos: No it's fine. The question is why you need duplicated new data at all? I..e 

```perl

my $x1 = { old => 'data' };
my $x2 = dclone($x1);
$x2->{new} = 'data2';

instead of 

my $y1 = { old => 'data' };
my $y2 = {};
$y2->{new} = 'data2';

```

In the last case there is no need of having duplicate data.

lib/Quant/Framework/VolSurface.pm
junbon: Yes, I remembered sending an email about this change in behaviour.

I honestly think that the check in

https://github.com/binary-com/perl-Quant-Framework/blob/master/lib/Quant/Framework/ImpliedRate.pm#L81

is bad. If the cache data is very old, we will use the wrong market data to
calculate price. Even though this will only happen in backtesting. We will
not know about it because everything behaves just fine but we are not
backtesting with the correct market data.

lib/Quant/Framework/TradingCalendar.pm
junbon: Yeah. Any concerns?

lib/Quant/Framework/VolSurface.pm
basiliscos: I agree that the sample code is not very good. But, please, follow it's logic. Would in all code be consistent behavior it will be much more easy to refactor all places.

Also afaik, in several places we provide `$for_date = now()`, i.e. not for back pricing. The additional load on DB would be undesirable.

lib/Quant/Framework/TradingCalendar.pm
basiliscos: Use `_cache` object per instance.

mm-binary: @basiliscos If we do this per-instance, then it won't be effective caching. I think this is not a big deal.

tfoertsch: Could you please call `time` only once?

```perl
my $t = time;
if ($t - $cache_init_time ...) {
    $cache_init_time = $t;
```

The point is `time` is a syscall which is much more expensive than just a variable.

junbon: noted.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -48,11 +46,74 @@ has chronicle_writer => (
...
 
+has document => (
+    is         => 'rw',
+    lazy_build => 1,
+);
+
+sub _build_document {
+    my $self = shift;
+
+    my $document = $self->chronicle_reader->get('volatility_surfaces', $self->symbol);
```
tfoertsch: Can `$document` be undef after this? I guess it can. If that happens, does it make sense to return an empty document `{}`? Something like:

```perl
my $document = ...;
$document = $self->chronicle_reader->get_for('volatility_surfaces', $self->symbol, $self->for_date->epoch)
    if $self->for_date and $self->for_date->epoch < Date::Utility->new($document->{date})->epoch;
    
return $document // {};
```

Does that make sense?

tfoertsch: Also, this is a lazy initializer. So, if you call `$obj->document` and the builder returns undef, Moose does not accept that as the value of the attribute. If you call `->document` in a loop, it calls the builder every time.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -48,11 +45,74 @@ has chronicle_writer => (
...
+    if ($self->for_date and $self->for_date->epoch < Date::Utility->new($document->{date})->epoch) {
```
tfoertsch: is this faster than [`is_before` / `is_after`](https://metacpan.org/pod/Date::Utility#is_before)

lib/Quant/Framework/VolSurface.pm
```diff
@@ -608,354 +459,79 @@ Returns the ask/bid spread for smile of this volatility surface
...
-
-    my $smile_spread;
-
-    if ($self->_market_name eq 'indices' and $self->type eq 'moneyness') {
-
+    my $surface = $self->surface;
+    # if a surface has a minimum volatility spread that needs to be applied,
+    # we will do the check.
+    if ($self->can('min_vol_spread')) {
+        my %market_points = map { $_ => 1 } @{$self->original_term_for_spread};
```
tfoertsch: There is a better way to do this. You don't use `map` to initialize the hash with boolean values. Instead you just make sure the hash elements exist and then use `exists`:

```
# perl -MBenchmark=:hireswallclock,timethese,cmpthese -e '
    my @exst=map {"a$_"} 1..50;
    my @ask=map {"a$_"} 1..100;
    my $sub1=sub {my %h=map {($_=>1)} @exst; $h{$ask[int rand 0+@ask]} for (1..20)};
    my $sub2=sub {my %h; @h{@exst}=(); exists $h{$ask[int rand 0+@ask]} for (1..20)};
    cmpthese timethese -2, {map=>$sub1, exist=>$sub2}
'
Benchmark: running exist, map for at least 2 CPU seconds...
     exist: 2.15253 wallclock secs ( 2.15 usr +  0.00 sys =  2.15 CPU) @ 100018.60/s (n=215040)
       map: 2.01079 wallclock secs ( 2.01 usr +  0.00 sys =  2.01 CPU) @ 52093.03/s (n=104707)
          Rate   map exist
map    52093/s    --  -48%
exist 100019/s   92%    --
```

This is surely a contrived example. But it shows the effect.

lib/Quant/Framework/VolSurface.pm
junbon: Thanks for noting. I think I would die if ```$document``` is undefined.

lib/Quant/Framework/VolSurface.pm
junbon: noted.

lib/Quant/Framework/VolSurface.pm
junbon: Slightly but not significant.

tfoertsch:
as discussed in person, please think about replacing the HoH data structure by AoA.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -983,510 +554,373 @@ sub get_day_for_tenor {
...
-    if (!isnum($sought_point)) {
-        die("Point[$sought_point] must be a number");
-    }
+    my @validation_methods = $self->_validation_methods;
 
-    if ($day <= 0 or $sought_point < 0) {
-        die("get_volatility requires positive numeric days[$day] and sought_point[$sought_point]");
+    while (not $self->validation_error and my $method = shift @validation_methods) {
+        try { $self->$method } catch { $self->validation_error($_) };
     }
```
tfoertsch: This loop is better done like this:

```perl
try {
    $self->$_ for ($self->_validation_methods);
}
catch {
    $self->validation_error($_);
}
```

The point is setting up an `eval` block in perl is relatively expensive. You want to bail out on the first error anyway. So, it's enough to setup the `eval` block once and catch the first error.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -608,354 +455,78 @@ Returns the ask/bid spread for smile of this volatility surface
...
+    my ($self, $seek, $available_points) = @_;
+    die('Need 2 or more term structures to interpolate.')
+        if scalar @$available_points <= 1;
 
+    return @{find_closest_numbers_around($seek, $available_points, 2)};
 }
 
-sub _compute_and_set_smile_spread {
-    my ($self, $day) = @_;
+sub _is_between {
```
evostrov: isn't better to move this method from this module?

lib/Quant/Framework/VolSurface.pm
```diff
@@ -983,510 +554,373 @@ sub get_day_for_tenor {
...
-    die("day[$day] must be a number.")
-        unless looks_like_number($day);
+sub _validate_volatility_jumps {
+    my $self = shift;
 
-    my $surface          = $self->surface;
-    my $atm_spread_point = $self->atm_spread_point;
-    my $smile            = $args->{smile};
-    my $vol_spread       = $self->get_smile_spread($day);
+    my $existing = $self->meta->name->new({
```
tfoertsch: It's usually a sign of bad design if you have to refer to Moose introspection techniques. What are you trying to do here?

Can `__PACKAGE__->new` or `ref($self)->new` help?

lib/Quant/Framework/VolSurface.pm
```diff
@@ -983,510 +554,373 @@ sub get_day_for_tenor {
...
 
-    $surface->{$day}->{smile} = $smile;
+    my @points = @{$self->smile_points};
+    my $type   = $self->type;
 
-    if ($atm_spread_point eq 'atm_spread') {
-        $surface->{$day}->{atm_spread} = $vol_spread->{'atm_spread'};
-    } else {
-        $surface->{$day}->{vol_spread} = $vol_spread->{'vol_spread'};
+    for (my $i = 1; $i < $#new_expiry; $i++) {
```
tfoertsch: You are skipping the first and last element. I guess you are aware of that and it's on purpose.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -983,510 +554,373 @@ sub get_day_for_tenor {
...
+                to    => $new_expiry[$i],
+            });
+            my $existing_vol = $existing->get_volatility({
+                $type => $sought_point,
+                from  => $existing->recorded_date,
+                to    => $existing_expiry[$i],
+            });
+            my $diff            = abs($new_vol - $existing_vol);
+            my $percentage_diff = $diff / $existing_vol * 100;
+            if ($diff > 0.03 and $percentage_diff > 100) {
```
tfoertsch: Can we assume that vols are positive numbers or 0?

If so, this condition can be  reduced to `$diff > 0.03 and $diff > $existing_vol`, right? Then you can move the `$percentage_diff` calculation inside the `if`.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -983,510 +554,373 @@ sub get_day_for_tenor {
...
-
-        $current_date = $vdoc->{date};
-
-        push @dates, $current_date;
+    if (not exists $clone_args{surface_data}) {
+        $clone_args{surface_data} = {%{$self->surface_data}};
     }
 
-    return \@dates;
+    return $self->meta->name->new(\%clone_args);
```
tfoertsch: see my remark about Moose introspection above.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -983,510 +554,373 @@ sub get_day_for_tenor {
...
-        Quant::Framework::VolSurface::Validator->new->validate_surface($self);
-        $err = 'Surface has smile flags: ' . $self->get_smile_flags
-            if $self->get_smile_flags;
+    # For backward compatibility in volatility spread,
+    # we will convert the legacy structure to the new structure here.
+    foreach my $tenor (keys %$surface) {
+        if (exists $surface->{$tenor}{atm_spread}) {
+            my $spread = delete $surface->{$tenor}{atm_spread};
+            $surface->{$tenor}{vol_spread} = {$self->atm_spread_point => $spread};
+        }
```
tfoertsch: This is only compatibility code. Still it hurts the eye.

```perl
for my $v (values %$surface) {
    defined($_ = delete $v->{$tenor}->{atm_spread}) and
        $v->{$tenor}->{vol_spread} = {$self->atm_spread_point => $_};
}
```

something like this.

lib/Quant/Framework/VolSurface.pm
junbon: noted.

lib/Quant/Framework/VolSurface.pm
junbon: move it to?

lib/Quant/Framework/VolSurface.pm
junbon: I can live with __PACKAGE__->new

lib/Quant/Framework/VolSurface.pm
junbon: I somehow feel that mine is easier to read. Will keep this.

lib/Quant/Framework/VolSurface.pm
junbon: Yes. Volatility will not be negative. I will change that.

lib/Quant/Framework/VolSurface.pm
junbon: I didn't mean to ignore the last smile. Will change the condition to <=

lib/Quant/Framework/VolSurface.pm
```diff
@@ -983,510 +554,375 @@ sub get_day_for_tenor {
...
-    $surface->{$day}->{smile} = $smile;
+    my @points = @{$self->smile_points};
+    my $type   = $self->type;
 
-    if ($atm_spread_point eq 'atm_spread') {
-        $surface->{$day}->{atm_spread} = $vol_spread->{'atm_spread'};
-    } else {
-        $surface->{$day}->{vol_spread} = $vol_spread->{'vol_spread'};
+    for (my $i = 1; $i <= $#new_expiry; $i++) {
+        for (my $j = 0; $j < $#points; $j++) {
```
mm-binary: You can also use:
`for my $sought_point ($points)`

jy-binary:
conflicts

PR #50, opened 2016-07-18
initial files

mm-binary:


lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,261 @@
...
+feed db if it is not present in cache.
+
+=cut
+
+sub spot_tick {
+    my $self = shift;
+
+    return $self->tick_at($self->for_date->epoch, {allow_inconsistent => 1}) if $self->for_date;
+
+    my $value = Cache::RedisDB->get('COMBINED_REALTIME', $self->underlying_config->symbol);
```
jy-binary: no combined

lib/Quant/Framework/Spot/DatabaseAPI.pm
```diff
@@ -0,0 +1,653 @@
+package Quant::Framework::Spot::DatabaseAPI;
+
+=head1 NAME
+
+Quant::Framework::Spot::DatabaseAPI
+
+=head1 DESCRIPTION
+
+The API class with which we can query feed/otn database.
```
jy-binary: otn??

lib/Quant/Framework/Spot/DatabaseAPI.pm
```diff
@@ -0,0 +1,653 @@
...
+
+sub _query_ticks {
+    my $self      = shift;
+    my $statement = shift;
+
+    my $symbol = $self->underlying;
+
+    my @ticks;
+    if ($statement->execute()) {
+        my ($epoch, $quote, $runbet_quote, $bid, $ask);
```
jy-binary: runbetquote???

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,266 @@
...
+    return $self->pipsized_value($last_price);
+}
+
+=head2 pipsized_value
+
+Resize a value to conform to the pip size of this underlying
+
+=cut
+
+sub pipsized_value {
```
junbon: Shouldn't this be in ```Underlying.pm``` rather than here? Pip sizing should only be in client facing end.

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,266 @@
...
+use Quant::Framework::Spot::Tick;
+use Quant::Framework::Spot::DatabaseAPI;
+
+=head1 NAME
+
+Quant::Framework::Spot - Used to store/retrieve spot prices into/from local Redis storage
+
+=cut
+
+has for_date => (
```
junbon: We still keep for_date implementation?

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,266 @@
...
+    is => 'ro',
+);
+
+has underlying_config => (
```
junbon: Isn't this supposed to be a required attribute?

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,266 @@
...
+=head2 tick_at
+
+This method receives a timestamp and an option for inconsistency.
+Returns the tick for current symbol at the given time. If inconssitency is enabled
+and there is no tick at that moment, the latest tick before that time is 
+returned.
+
+=cut
+
+sub tick_at {
```
junbon: Can we not have something like this:

```
my $current_tick = Cache::RedisDB->get($self->default_redis_key, $self->underlying_config->symbol);

if ($timestamp < $current_tick->{epoch}) {
    #fetch and return tick from database
} else {
    # return $current_tick
}

```
Then we can remove ```sub spot_*``` because they just look very confusing here. They should be in ```Underlying.pm```

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,266 @@
...
+    is => 'ro',
+);
+
+has feed_api => (
+    is => 'ro',
+);
+
+has default_redis_key => (
+    is      => 'ro',
+    default => 'COMBINED_REALTIME',
```
jy-binary: can you change to 'REALTIME' or 'QUOTES' ? The word "combined" makes little sense in this module.

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,266 @@
...
+    my $self = shift;
+
+    my $tick = $self->spot_tick;
+
+    return ($tick) ? $tick->as_hash : undef;
+}
+
+=head2 tick_at
+
+This method receives a timestamp and an option for inconsistency.
```
jy-binary: what does 'inconsistency' mean? Please explain and document the concept.

lib/Quant/Framework/Spot/DatabaseAPI.pm
```diff
@@ -0,0 +1,673 @@
...
+    my ($self, %args) = @_;
+
+    my $underlying    = $args{underlying};
+    my $system_symbol = $args{system_symbol};
+    my $pipsize       = $args{pip_size};
+    my $start_time    = Date::Utility->new($args{start_time})->db_timestamp;
+    my $end_time      = Date::Utility->new($args{end_time} // time)->db_timestamp;
+    my @sql_args      = ($system_symbol, $start_time, $end_time);
+    my $sql =
+        'SELECT EXTRACT(epoch FROM ts) AS epoch, spot FROM ' . ($self->default_prefix // '') . 'tick WHERE underlying = $1 AND ts >= $2 AND ts <= $3';
```
jy-binary: How is a module user supposed to know how to set up the DB? Is there any documentation somewhere explaining the DB schema format?

lib/Quant/Framework/Spot/DatabaseAPI.pm
```diff
@@ -0,0 +1,673 @@
...
+
+get ticks from feed db filtered by
+    - start_time, end_time, limit - all ticks between <start_time> and <end_time> and limit to <limit> entries.
+
+Returns
+     ArrayRef[Quant::Framework::Spot::Tick]
+
+=cut
+
+sub ticks_start_end_with_limit_for_charting {
```
jy-binary: why do we need special routines "for charting"? What is different when we need charts? Please explain in the comments.

lib/Quant/Framework/Spot/DatabaseAPI.pm
mm-binary: I have explained about the feed.tick table and it's expected structure in the README file of the repo. Will review the documentation and make sure it's complete.

lib/Quant/Framework/Spot.pm
mm-binary: I think we discussed this and the conclusion was to keep this one, un-touched.

lib/Quant/Framework/Spot.pm
mm-binary: We need pipsized_value spot quote in `Validator` and `Delta` modules.

lib/Quant/Framework/Spot/Tick.pm
```diff
@@ -0,0 +1,124 @@
...
+    return +{map { $_ => $self->$_ } grep { defined $self->$_ } qw(epoch quote symbol bid ask)};
+}
+
+__PACKAGE__->meta->make_immutable;
+
+1;
+
+=head1 LICENSE AND COPYRIGHT
+
+Copyright 2012 RMG Technology (M) Sdn Bhd
```
jy-binary: remove all copyright notices. This is open-source code, there is no copyright.

.travis.yml
```diff
@@ -4,6 +4,8 @@ perl:
     - "5.22"
     - "5.18"
     - "5.14"
+services:
+    - redis-server
```
jy-binary: the code is calling SQL but I'm not seeing any postgres database set up in the tests?

mm-binary: I think it is provided by default by travis environment because tests were all green. But I will add the dependency here to be more clear.

lib/Quant/Framework/Spot/Tick.pm
mm-binary: Done.

lib/Quant/Framework/Spot.pm
mm-binary: It is already explained in the README file. I added the same explanation to the POD.

lib/Quant/Framework/Spot.pm
mm-binary: Yes. Fixed.

lib/Quant/Framework/Spot.pm
mm-binary: I believe this might cause trouble in operation. Because in case of a delay in the feed for any reason, there will be a lot of DB requests which may cause issue for the database server. Moreover, we have a `for_date` parameter which we can use to determine which source we need to consult. So why not use it?

lib/Quant/Framework/Spot.pm
junbon: No we don't need pip sized value in calculation. @rakshit-binary clarify?

rakshit-binary: Agree with @junbon. Pipsized values aren't needed for QF calculations.

lib/Quant/Framework/Spot.pm
rakshit-binary: what is it used for ?

mm-binary: This is used to indicate the caller wants information at a specific point in time (not the latest data).

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,250 @@
...
+    return $self->tick_at($self->for_date->epoch, {allow_inconsistent => 1}) if $self->for_date;
+
+    my $value = Cache::RedisDB->get($self->default_redis_key, $self->underlying_config->symbol);
+    my $tick;
+    if ($value) {
+        $tick = Quant::Framework::Spot::Tick->new($value);
+    } else {
+        $tick = $self->tick_at(time, {allow_inconsistent => 1});
+        if ($tick) {
+            $self->set_spot_tick($tick);
```
basiliscos: Why do you need that side-effect? Why this side effect isn't used when we get the value from Redis (#76)

If you'll need that side-effect, it should be mentioned in th doc above

README.md
```diff
@@ -493,3 +493,106 @@ my $vol = $surface->get_volatility(
 my $utc21_surface = $surface->generate_surface_for_cutoff('UTC 21:00');
 
 ```
+##Quant::Framework::Spot
+
+This module is responsible for storage and retrieval of spot prices for different underlyings. The spot price is stored in local Redis for fast retrieval. For historical data, a database handle is expected as input which will be used to fetch historical spot and OHLC values. 
+
+There are some naming conventions in this module which user of the module needs to obey in order to use provided utilities. These conventions determine naming of tables and functions in the database. The tables and their structure are designed to match general layout of database tables which store ticks. For example a table `feed.tick` is expected to be in the database where it contains underlying name, timestamp and spot value. 
+
```
basiliscos:
```perl
my $statement = $self->dbh->prepare_cached('SELECT * FROM ohlc_start_end($1, $2, $3, $4, $5)', {}, 3);

```

I think there are more assumptions, like the certain DB-procedures should be present in DBMS-level, as well as DBMS should support them at all.

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,250 @@
...
+=cut
+
+sub closing_tick_on {
+    my ($self, $end, $closing) = @_;
+    my $date = Date::Utility->new($end);
+
+    if ($closing and time > $closing->epoch) {
+        my $ohlc = $self->feed_api->ohlc_start_end({
+            start_time         => $date,
+            end_time           => $date,
```
basiliscos: Usually, if `stat` and `end` are the same, the aggregation result is zero.

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,250 @@
...
+            aggregation_period => 86400,
```
basiliscos: @tfoertsch, @mm-binary  do we fetch all OHLC ticks for 24-hours, only to get the price of the last ticket? Isn't this over-kill?

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,250 @@
...
+        for_date => Date::Utility->new('2016-09-11 12:29:10')
+    });
+
+    ...
+
+=head1 SUBROUTINES/METHODS
+
+=head2 $self->spot_tick
+
+Get last tick value for symbol from Redis. It will rebuild value from
```
basiliscos: > from Redis

or from DB

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,250 @@
...
+sub set_spot_tick {
+    my ($self, $value) = @_;
+
+    my $tick;
+    if (ref $value eq 'Quant::Framework::Spot::Tick') {
+        $tick  = $value;
+        $value = $value->as_hash;
+    } else {
+        $tick = Quant::Framework::Spot::Tick->new($value);
+    }
```
basiliscos: Can we get rid of this if..else at all? I.e. either we expect `Quant::Framework::Spot::Tick` or hash on input. Not both. 

```perl
$tick = Quant::Framework::Spot::Tick->new($value);
```

you create `$tick` object, but never use it later.

lib/Quant/Framework/Spot.pm
```diff
@@ -0,0 +1,250 @@
...
+    my $last_tick = $self->spot_tick;
+    $last_price = $last_tick->quote if $last_tick;
+
+    return $last_price;
+}
+
+
+=head2 spot_time
+
+The epoch timestamp of the latest recorded tick in the .realtime file or undef if we can't find one.
```
basiliscos: > tick in the .realtime file

`.realtime file` ? What is it?

lib/Quant/Framework/Spot/DatabaseAPI.pm
```diff
@@ -0,0 +1,663 @@
...
+sub ticks_start_end {
+    my $self = shift;
+    my $args = shift;
+
+    my $start_time;
+    my $end_time;
+    $start_time = Date::Utility->new($args->{start_time})->datetime_yyyymmdd_hhmmss
+        if ($args->{start_time});
+    $end_time = Date::Utility->new($args->{end_time})->datetime_yyyymmdd_hhmmss
+        if ($args->{end_time});
```
basiliscos: I think `$args->{start_time}` and `$args->{end_time}` are mandatory arguments here?

lib/Quant/Framework/Spot/DatabaseAPI.pm
basiliscos: `tick_at_for_interval($1, $2, $3, $4)` - that stored procedure also belongs to the DB-level assumptions.

basiliscos: Also it assumes the `ts` table presense

basiliscos: `EXTRACT(epoch FROM ts)` - I think this is PG-specific request. It will not run, for example, on MySQL

lib/Quant/Framework/Spot/DatabaseAPI.pm
```diff
@@ -0,0 +1,663 @@
...
+    $statement->bind_param(2, $start_time);
+    $statement->bind_param(3, $end_time);
+    $statement->bind_param(4, $self->use_official_ohlc ? 'TRUE' : 'FALSE');
+
+    return $self->_query_ohlc($statement);
+}
+
+=head2 combined_realtime_tick
+
+Returns combined tick for the given start/end time period.
```
basiliscos: What this method does? What is the combined realtime tick and how it differs from the rest of ticks? 

 What is the combined realtime tick *for the given period*? If I'll specify 1 year period, will it return the the tick of 1st of January, or the 31 of December or somewhere in July?

lib/Quant/Framework/Spot/DatabaseAPI.pm
```diff
@@ -0,0 +1,663 @@
...
+
+        while ($statement->fetch()) {
+            my $tick_compiled = Quant::Framework::Spot::Tick->new({
+                symbol => $symbol,
+                epoch  => $epoch,
+                quote  => $quote,
+                bid    => $bid,
+                ask    => $ask,
+            });
+            $tick_compiled->invert_values if ($self->invert_values);
```
basiliscos: Can we remove the check `if ($self->invert_values)` at all? I propose to have method `inverted` in `Quant::Framework::Spot::Tick` that will do that on demand.  

It would be more clear and faster, than do checks like that spread among the code.

lib/Quant/Framework/Spot/Tick.pm
```diff
@@ -0,0 +1,119 @@
...
+
+sub invert_values {
+    my $self = shift;
+
+    $self->quote(1 / $self->quote);
+    $self->bid(1 / $self->bid) if $self->bid;
+    $self->ask(1 / $self->ask) if $self->ask;
+
+    return $self;
+}
```
basiliscos: Instead of inverting values, I propose to return inverted tick. It don't have side-effects.

README.md
basiliscos: > Cache::RedisDB

Another assumption, that we have running redis at some *predefined* port and host. I think this nowhere is possible to configure them, so, that should be stated.

lib/Quant/Framework/Spot/Tick.pm
mm-binary: I prefer not to. Because in some places in the code, this method is used without catching output value. Also the name of the methods implies it changes the status of the object.

lib/Quant/Framework/Spot/DatabaseAPI.pm
mm-binary: Fixed.

lib/Quant/Framework/Spot.pm
mm-binary: Fixed.

lib/Quant/Framework/Spot.pm
mm-binary: We need `$tick` for return value of the method.

lib/Quant/Framework/Spot.pm
mm-binary: Diben.

lib/Quant/Framework/Spot.pm
mm-binary: Added.

README.md
mm-binary: We are using default config of Cache::RedisDB which is localhost with default port. I will add this to the documentation.

README.md
```diff
@@ -493,3 +493,123 @@ my $vol = $surface->get_volatility(
...
+my $ohlc = $db->ohlc_start_end({
+    start_time  => $start_ts,
+    end_time    => $end_ts,
+    aggregation_period => $num_seconds_per_ohlc_instance
+});
+
+my @ohlc_bars = @{$ohlc};
+```
+
+The first call above, request for latest tick at given time (`$timestamp`) and the second one askas
```
evostrov: typo `askas`

lib/Quant/Framework/Spot/DatabaseAPI.pm
mm-binary: Seems this is not used anywhere. I removed the method

README.md
mm-binary: Fixed.

lib/Quant/Framework/Spot.pm
mm-binary: Ok.

lib/Quant/Framework/Spot/DatabaseAPI.pm
mm-binary: Documentation is updated accordingly.

lib/Quant/Framework/Spot/DatabaseAPI.pm
mm-binary: Added.

README.md
```diff
@@ -493,3 +493,122 @@ my $vol = $surface->get_volatility(
 my $utc21_surface = $surface->generate_surface_for_cutoff('UTC 21:00');
 
 ```
+##Quant::Framework::Spot
```
jy-binary: Nice :-) I like the new documentation

README.md
```diff
@@ -493,3 +493,122 @@ my $vol = $surface->get_volatility(
...
+my @ohlc_bars = @{$ohlc};
+```
+
+The first call above, request for latest tick at given time (`$timestamp`) and the second one asks
+for daily OHLC values at given time.
+
+`allow_inconsistent` parameters determines how the case should be handled when there is no tick at the 
+exact requested timestamp. If value of this parameter is `1`, then the code will return tick at or before the given timestamp.
+If it is passed as `0`, the code will return any tick at or after given timestamp.
+
```
yngshan: It will return tick at the epoch with required one tick after that

https://github.com/regentmarkets/bom-postgres-feeddb/blob/master/config/sql/schema_7_up.sql#L19

lib/Quant/Framework/Spot/DatabaseAPI.pm
```diff
@@ -0,0 +1,626 @@
...
+}
+
+=head2 ohlc_daily_until_now_for_charting
+
+Returns daily OHLC values for the period starting from given date till today.
+
+=cut
+
+sub ohlc_daily_until_now_for_charting {
+    my ($self, $args) = @_;
```
yngshan: Just a note for future refactoring: ohlc_daily_until_now_for_charting and ohlc_start_end_with_limit_for_charting only used for BO which most likely no longer valid. We might need to remove them in future

README.md
mm-binary: Documentation is updated accordingly. Thanks.

PR #53, opened 2016-07-26
Remove default_interest_rate

evostrov:


lib/Quant/Framework/InterestRate.pm
```diff
@@ -130,9 +130,10 @@ sub interest_rate_for {
     my ($self, $tiy) = @_;
 
     # timeinyears cannot be undef
-    $tiy ||= 0;
+    $tiy //= 0;
```
tfoertsch: If an empty string is passed as `$tiy`, the previous version would have converted it silently into `0`. Not sure what this version does. Anyway, I think this is minor.

evostrov: sure you are right, I'll rollback my changes

PR #54, opened 2016-07-27
Refactoring Dividend

shahrizal-shaari:


junbon:
Did we remove default dividend in this PR?

shahrizal-shaari:
@junbon No. I tried to do it together, but looks complicated to do it in one card. So decided to do it in separate card.

lib/Quant/Framework/TradingCalendar.pm
```diff
@@ -1305,7 +1305,7 @@ sub _build_asset {
     return unless $self->underlying_config->asset_symbol;
     my $type = $self->underlying_config->asset_class;
 
-    my $which = $type eq 'currency' ? 'Quant::Framework::Currency' : 'Quant::Framework::Asset';
+    my $which = $type eq 'currency' ? 'Quant::Framework::Currency' : 'Quant::Framework::Dividend';
```
junbon: I am not quite sure about the name for these modules. We used to have Currency and Asset as a breakdown of an instrument. Since the name dividend does not apply to market like Forex, should we have ```QF::Asset``` and ```QF::QoutedCurrency```. Then we could have ```QF::Asset->new->rate_for($x)``` and ```QF::QuotedCurrency->new->rate_for($y)```. Not sure which makes more sense in Financial world. 

@rakshit-binary input?

lib/Quant/Framework/Dividend.pm
```diff
@@ -201,14 +201,15 @@ sub dividend_rate_for {
...
+        ? Quant::Framework::Currency->new({
+            symbol           => $self->underlying_config->asset_symbol,
+            for_date         => $self->for_date,
+            chronicle_reader => $self->chronicle_reader,
+            chronicle_writer => $self->chronicle_writer,
+        })
+        : $self;
 
     if ($self->underlying_config->uses_implied_rate_for_asset) {
         $rate = $asset->rate_implied_from($self->underlying_config->rate_to_imply_from, $tiy);
```
mm-binary: `$asset` can be `$self` which is a `Dividend` module while we don't have `rate_implied_from` in dividend.

lib/Quant/Framework/Utils/Builder.pm
```diff
@@ -153,7 +152,7 @@ sub build_asset {
     return unless $self->underlying_config->asset_symbol;
     my $type = $self->underlying_config->asset_class;
 
-    my $which = $type eq 'currency' ? 'Quant::Framework::Currency' : 'Quant::Framework::Asset';
+    my $which = $type eq 'currency' ? 'Quant::Framework::Currency' : 'Quant::Framework::Dividend';
```
mm-binary: We have both `build_currency` and `build_dividend`. So maybe we can remove `build_asset`

mm-binary: Just noted that we are passing `asset_symbol` while in `build_dividend` we use `symbol`. so I think it's better to keep this one.

lib/Quant/Framework/TradingCalendar.pm
rakshit-binary: FX - rate for Asset
Commodities - this is a substitute for cost of carry
Indices - this is a substitute for dividends (making it as a continuous rate).
Stocks - dividends are handled via Corporate Action

lib/Quant/Framework/Dividend.pm
rakshit-binary: this is go away, and we will only have Asset.pm

lib/Quant/Framework/Asset.pm
```diff
@@ -52,26 +62,246 @@ has chronicle_writer => (
...
+
+    my $document = $self->chronicle_reader->get('dividends', $self->symbol);
+
+    if ($self->for_date and $self->for_date->epoch < Date::Utility->new($document->{date})->epoch) {
+        $document = $self->chronicle_reader->get_for('dividends', $self->symbol, $self->for_date->epoch);
+
+        # This works around a problem with Volatility surfaces and negative dates to expiry.
+        # We have to use the oldest available surface.. and we don't really know when it
+        # was relative to where we are now.. so just say it's from the requested day.
+        # We do not allow saving of historical surfaces, so this should be fine.
```
yngshan: This comment seem irrelevant to dividend :)

shahrizal-shaari:
Will recreate a new one to resolve conflict.

PR #56, opened 2016-08-05
Shahrizal/qf refactor dividend

shahrizal-shaari:


lib/Quant/Framework/Asset.pm
```diff
@@ -2,46 +2,56 @@ package Quant::Framework::Asset;
 
 =head1 NAME
 
-Quant::Framework::Asset
+Quant::Framework::Asset - A module to save/load dividends
```
mm-binary: If this module is only for dividends, so why it's name is not Dividend?

lib/Quant/Framework/Asset.pm
```diff
@@ -52,26 +62,241 @@ has chronicle_writer => (
...
+
+has recorded_date => (
+    is         => 'ro',
+    isa        => 'Date::Utility',
+    lazy_build => 1,
+);
+
+sub _build_recorded_date {
+    my $self = shift;
+    return Date::Utility->new($self->document->{date});
```
mm-binary: If the document is new (there is no existing document), `$self->document->{date}` will throw error. Because of this, we need to add 3 lines of code to the `save` method to check for this and insert an empty document just to handle this exception. Can you handle it here and remove those lines from save?

lib/Quant/Framework/Asset.pm
mm-binary: In `dividend_rate_for` I see that this module is doing something with `Currency` too. So maybe you can modify this description.

lib/Quant/Framework/Asset.pm
```diff
@@ -52,26 +62,241 @@ has chronicle_writer => (
...
+            chronicle_writer  => $self->chronicle_writer
+        });
+
+        #TODO: rewrite this using an instance of InterestRate
+        my $r_rate = $ir->interest_rate_for($duration_in_year);
+
+        my $adj_present_value = $adjustment * exp(-$r_rate * $duration_in_year);
+        my $s_adj = ($duration_in_sec - $sec_away_from_action) / ($duration_in_sec) * $adj_present_value;
+        $dS -= $s_adj;
+        my $k_adj = $sec_away_from_action / ($duration_in_sec) * $adj_present_value;
```
mm-binary: This expression is ambiguous. can you add paranthesis?

PR #59, opened 2016-08-11
junbon/min_vol_spread

junbon:


lib/Quant/Framework/VolSurface.pm
```diff
@@ -167,11 +167,21 @@ sub _build_surface {
...
             warn("Unknown tenor[$maturity] found on volatility surface for " . $self->symbol);
+            next;
+        }
+
+        #check and add min_vol_spread for shorter term vol_spreads
+        next if (not exists $surface_data{$maturity}{vol_spread} or $maturity >= 30);
+        foreach my $point (keys %{$surface_data{$maturity}{vol_spread}}) {
+            if ($self->can('min_vol_spread') and $surface_data{$maturity}{vol_spread}{$point} < $self->min_vol_spread) {
+                $surface_data{$maturity}{vol_spread}{$point} = $self->min_vol_spread;
+            }
```
yngshan: What is the difference between this check and the one at https://github.com/junbon/perl-Quant-Framework/blob/a2a20d01e4033f1f20c9e268c9058517e7860a44/lib/Quant/Framework/VolSurface.pm#L497


They are same right

yngshan: @junbon , in the old code, we add the min vol spread. In your code, you are set it to min vol spread. I think we want to add

lib/Quant/Framework/VolSurface.pm
```diff
@@ -418,60 +419,34 @@ sub get_spread {
...
-
-sub _get_max_spread_from_surface {
-    my ($self, $day) = @_;
-
     my $surface = $self->surface;
-    if (exists $surface->{$day}) {
-        return max(values %{$surface->{$day}{vol_spread}});
+    # prevent autovivification
+    my $smile_spread = (exists $surface->{$day} and exists $surface->{$day}{vol_spread}) ? $surface->{$day}{vol_spread} : +{};
+    $smile_spread = $self->get_smile_spread($day) unless keys %$smile_spread;
```
basiliscos: I don't think that you need `keys` here. I.e. 

```perl
$smile_spread = $self->get_smile_spread($day) unless %$smile_spread;
```

should be enough

lib/Quant/Framework/VolSurface.pm
```diff
@@ -418,60 +419,34 @@ sub get_spread {
...
+    my $spread;
+    if ($sought_point eq 'atm') {
+        $spread = $smile_spread->{$self->atm_spread_point};
+    } elsif ($sought_point eq 'max') {
+        $spread = max(values %$smile_spread);
     }
 
-    my $smile_spread = $self->get_smile_spread($day);
-    return max(values %$smile_spread);
+    die 'Unrecognized spread type ' . $sought_point unless $spread;
```
basiliscos: I suggest to have `else` branch above, and insert `die` there.

lib/Quant/Framework/VolSurface.pm
```diff
@@ -418,60 +419,34 @@ sub get_spread {
...
+    return $spread + $self->min_vol_spread if ($self->can('min_vol_spread') and $day < 30 and $spread < $self->min_vol_spread);
```
basiliscos: That instruction is not obvious for me. I expected something like

```perl
return max($self->min_vol_spread, $spread);
```

Why you add min vol spread?

junbon: This is where the hell break lose. We actually want to add min_vol_spread to the raw spread.

basiliscos: Yes, I've noted that it was the same previous implementation too.

PR #63, opened 2016-09-01
junbon/zero_duration

junbon:
Handles zero duration better. This happens in proposal open contract at expiry when database does not have the exit tick populated.

### DIFF
--- a/lib/Quant/Framework/Spot.pm
+++ b/lib/Quant/Framework/Spot.pm
@@ -166,8 +166,7 @@ sub closing_tick_on {
         aggregation_period => 86400,
     });
 
-    if ($ohlc and scalar @{$ohlc} > 0) {
-
+    if (@$ohlc > 0) {
         # We need a tick, but we can only get an OHLC
         # The epochs for these are set to be the START of the period.
         # So we also need to change it to the closing time. Meh.
@@ -179,7 +178,7 @@ sub closing_tick_on {
         });
     }
 
-    return;
+    return undef;
 }
 
 =head2 $self->set_spot_tick($value)
--- a/lib/Quant/Framework/Spot.pm
+++ b/lib/Quant/Framework/Spot.pm
@@ -158,7 +158,10 @@ sub closing_tick_on {
     $date = Date::Utility->new($date);
 
     my $closing = $self->calendar->closing_on($date);
-    return if not $closing or time <= $closing->epoch;
+    if (not $closing or time <= $closing->epoch) {
+        ##no critic (ProhibitExplicitReturnUndef)
+        return undef;
+    }
 
     my $ohlc = $self->feed_api->ohlc_start_end({
         start_time         => $date,
@@ -178,7 +181,8 @@ sub closing_tick_on {
         });
     }
 
-    return;
+    ##no critic (ProhibitExplicitReturnUndef)
+    return undef;
 }
 
 =head2 $self->set_spot_tick($value)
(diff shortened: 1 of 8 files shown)


lib/Quant/Framework/Spot.pm
```diff
@@ -152,29 +152,33 @@ Example : $underlying->closing_tick_on("10-Jan-00");
...
-                symbol => $self->underlying_config->symbol,
-                epoch  => $closing->epoch,
-                quote  => $not_tick->close,
-            });
-        }
+    die 'must pass in a date for closing_tick_on' unless $date;
+    $date = Date::Utility->new($date);
+
+    my $closing = $self->calendar->closing_on($date);
+    return if not $closing or time <= $closing->epoch;
```
binarytom: If it's valid to have no tick for the requested parameters, presumably we should return `undef` here? Our code might always be using scalar context, but it's easier to enforce it in the return than having to audit all callers.

If that's unexpected, then an exception might be more appropriate (since we're potentially throwing them anyway when passed bad parameters).

lib/Quant/Framework/Spot.pm
```diff
@@ -152,29 +152,33 @@ Example : $underlying->closing_tick_on("10-Jan-00");
...
+
+    my $ohlc = $self->feed_api->ohlc_start_end({
+        start_time         => $date,
+        end_time           => $date,
+        aggregation_period => 86400,
+    });
+
+    if ($ohlc and scalar @{$ohlc} > 0) {
```
binarytom: What are the implications of `$ohlc` being false here - feed problem? market is closed on `$date`? It's not clear to me why there's a distinction between empty `@$ohlc` and false `$ohlc`, does the latter indicate an error, and if so are we logging it?

lib/Quant/Framework/Spot.pm
junbon: Noted. Going to return undef.

lib/Quant/Framework/Spot.pm
junbon: I don't think there's no reason to check for $ohlc being false. Should be a
legacy code and got copied over. I will make the appropriate change.

lib/Quant/Framework/Spot.pm
```diff
@@ -152,30 +152,37 @@ Example : $underlying->closing_tick_on("10-Jan-00");
...
-                epoch  => $closing->epoch,
-                quote  => $not_tick->close,
-            });
-        }
+    die 'must pass in a date for closing_tick_on' unless $date;
+    $date = Date::Utility->new($date);
+
+    my $closing = $self->calendar->closing_on($date);
+    if (not $closing or time <= $closing->epoch) {
+        ##no critic (ProhibitExplicitReturnUndef)
```
mm-binary: Maybe you can add a `warn` here.

mm-binary:
LGTM

lib/Quant/Framework/Spot.pm
junbon: It will just clutter the log. I don't think it is necessary.

lib/Quant/Framework/VolSurface/Moneyness.pm
```diff
@@ -113,7 +113,15 @@ sub get_volatility {
...
 
-    die "Argument 'days' must be positive, non-zero number." if $internal_args{days} <= 0;
+    if ($internal_args{days} <= 0) {
+        $self->validation_error('Invalid request for get volatility. Surface recorded date ['
+                . $self->recorded_date->datetime
+                . '] requested period ['
+                . $args->{from}->datetime . ' to '
+                . $args->{to}->datetime
+                . ']');
+        return 0.01;    # return a 1% volatility but we will not sell on this volatility.
```
shahrizal-shaari: Why can't we return undef in case of error, instead of some value ?

